PR #123, opened 2013-09-11
Send text

Schmoozerd:
First commit is pretty clear:
- Remove MonsterText, MonsterYell, MonsterEmote, MonsterWhisper functions (int32 entry versions) and replace with a generic function MonsterText that operates on MangosLocaleString (which has the information about type, language)
- Pass the MangosLocaleString to the actual builder functions, to prevent a double lookup

The second commit:
The goal is to be able to handle sound (maybe later on emote) within the same "iterators" over the players then the actual strings. This should give some small performance benefits on the long run

(Both untested)

### DIFF
--- a/src/game/AchievementMgr.cpp
+++ b/src/game/AchievementMgr.cpp
@@ -67,6 +67,7 @@ namespace MaNGOS
                 data << uint8(0);
                 data << uint32(i_achievementId);
             }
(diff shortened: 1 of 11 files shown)


cyberium:
Abandonned

killerwife:
Maybe, but its needed. We are not able to send sounds and music.

PR #125, opened 2013-10-07
Improve - check compile

Schmoozerd:
Also any feedback is welcome.

Also poke @rsa - you said something about non-iso before - what exactly were you refering to?


cyberium:
Abandonned

PR #147, opened 2015-02-06
Add positive Effect spells for Arcane Bolt and Dummy Achievement Credit Auras.

Ulduar:
Arcane Bolt(Kalecgos/Kil'jaeden encounter) - on spellhit target - Initialize aggro Kil'jaeden on Kalecgos.
Achievement credit spells - also somehow cause aggro on the player. Bad, very bad.
Testing on Wotlk a long time ago.
https://github.com/cmangos/mangos-wotlk/blob/master/sql/updates/12434_01_mangos_spell_template.sql

### DIFF
--- a/src/game/SpellMgr.cpp
+++ b/src/game/SpellMgr.cpp
@@ -704,6 +704,7 @@ bool IsPositiveEffect(SpellEntry const* spellproto, SpellEffectIndex effIndex)
             switch (spellproto->Id)
             {
                 case 42436:                                 // Drink!
+                case 45666:                                 // Arcane Bolt
                 case 46650:                                 // Open Brutallus Back Door
                 case 62488:                                 // Activate Construct
                 case 64503:                                 // Water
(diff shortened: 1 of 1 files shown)


xfurry:
Why do you need positive target on the achiev completion spells?
Those spells are working just fine.

Ulduar:
First spell_template on Cataclysm is terribly (or rather not working) for Votlk works on the principle - I want to work, and do not want to - the disabled. Third, I have already described why I need a positive effect.

xfurry:
> First spell_template on Cataclysm is terribly (or rather not working)

On Cata spell_template is not working. I already explained that it would be tricky to make it work (not impossible) but I didn't have the time and motivation to do this.

> I have already described why I need a positive effect.

I am not aware of this. Please describe again.
The only thing that I know is that all of the achiev spells that I added here were tested and are working fine. So I can see no reason for this change.
If you are using YTDB, make sure that you have the achiev_criteria table updated.

Ulduar:
``` QUOTE
On Cata spell_template is not working. I already explained that it would be tricky to make it work (not impossible) but I didn't have the time and motivation to do this.
```
Ragnaros in Firelands has kill credit spell(normal difficulty), also achievement Ragna-roi has kill credit spell.
``` QUOTE
I am not aware of this. Please describe again.
```
The third attempt - do not know where and who tested it, but these spells initiate boss aggro on me.
``` QUOTE
If you are using YTDB, make sure that you have the achiev_criteria table updated.
```
This table has been filled with many more years ago. Much earlier than has appeared in SD2.
And indeed nothing to do with the table. Do you know the way that Algalon kill credit spell does not work? Do not count towards the achievement Observed.

xfurry:
Ok, then I will recheck all the Ulduar achievs spells, including Algalon and the issues that you reported.

PR #148, opened 2015-02-06
Add new command for flexible interaction.

Ulduar:
SCRIPT_COMMAND_RESPAWN_DELAY  for Rhonin and other - set respawn in db - full hack and 
not acceptable(sniff data for respawn)
SCRIPT_COMMAND_CORPSE_DELAY     for Patchwerk in Dk knight and Cenarius in quest 25551 Firelord.
Also interesting SCRIPT_COMMAND_WORLD_STATE and CONDITION_WORLD_STATE.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -319,3 +319,10 @@ Where "A  ->  B" means that the command is executed from A with B as target.
                                             * datalong = mailTemplateId
                                             * datalong2: AlternativeSenderEntry. Use as sender-Entry of the sent mail
                                             * dataint1: Delay (>= 0) in Seconds
+
+39 SCRIPT_COMMAND_CORPSE_DELAY              resultingSource = Creature
+                                            * datalong = corpse delay
+                                            
+40 SCRIPT_COMMAND_RESPAWN_DELAY              resultingSource = Creature
+                                            * datalong = respawn delay
+                                            
--- a/src/game/ScriptMgr.cpp
+++ b/src/game/ScriptMgr.cpp
@@ -724,6 +724,7 @@ void ScriptMgr::LoadScripts(ScriptMapMapName& scripts, const char* tablename)
             case SCRIPT_COMMAND_CORPSE_DELAY:              // 39
                 break;
             case SCRIPT_COMMAND_RESPAWN_DELAY              // 40
+                break;
             default:
             {
                 sLog.outErrorDb("Table `%s` unknown command %u, skipping.", tablename, tmp.command);
--- a/src/game/ScriptMgr.cpp
+++ b/src/game/ScriptMgr.cpp
@@ -721,6 +721,9 @@ void ScriptMgr::LoadScripts(ScriptMapMapName& scripts, const char* tablename)
                 }
                 break;
             }
+            case SCRIPT_COMMAND_CORPSE_DELAY:              // 39
+                break;
+            case SCRIPT_COMMAND_RESPAWN_DELAY              // 40
             default:
             {
                 sLog.outErrorDb("Table `%s` unknown command %u, skipping.", tablename, tmp.command);
@@ -1945,6 +1948,38 @@ bool ScriptAction::HandleScriptStep()
             uint32 deliverDelay = m_script->textId[0] > 0 ? (uint32)m_script->textId[0] : 0;
 
             MailDraft(m_script->sendMail.mailTemplateId).SendMailTo(static_cast<Player*>(pTarget), sender, MAIL_CHECK_MASK_HAS_BODY, deliverDelay);
+            break;
+        }
+        case SCRIPT_COMMAND_CORPSE_DELAY:                   // 39
+        {
+            // TODO - Remove this check after a while
+            if (pTarget && pTarget->GetTypeId() != TYPEID_UNIT && pSource && pSource->GetTypeId() == TYPEID_UNIT)
+            {
+                sLog.outErrorDb("DB-SCRIPTS: Process table `%s` id %u, command %u target must be creature, but (only) source is, use data_flags to fix", m_table, m_script->id, m_script->command);
+                pTarget = pSource;
+            }
+
+            if (LogIfNotCreature(pTarget))
+                break;
+
+            ((Creature*)pTarget)->SetCorpseDelay(m_script->corpse.corpseDelay);
+
+            break;
+        }
+        case SCRIPT_COMMAND_RESPAWN_DELAY:                   // 40
+        {
+            // TODO - Remove this check after a while
+            if (pTarget && pTarget->GetTypeId() != TYPEID_UNIT && pSource && pSource->GetTypeId() == TYPEID_UNIT)
+            {
+                sLog.outErrorDb("DB-SCRIPTS: Process table `%s` id %u, command %u target must be creature, but (only) source is, use data_flags to fix", m_table, m_script->id, m_script->command);
+                pTarget = pSource;
+            }
+
+            if (LogIfNotCreature(pTarget))
+                break;
+
+            ((Creature*)pTarget)->SetRespawnDelay(m_script->respawn.respawnDelay);
+            
             break;
         }
         default:
--- a/src/game/ScriptMgr.h
+++ b/src/game/ScriptMgr.h
@@ -118,7 +118,7 @@ enum ScriptCommand                                          // resSource, resTar
                                                             // datalong2: AlternativeSenderEntry. Use as sender-Entry
                                                             // dataint1: Delay (>= 0) in Seconds
     SCRIPT_COMMAND_CORPSE_DELAY             = 39,           // resSource Creature, datalong - corpse delay.
-    SCRIPT_COMMAND_SET_RESPAWN_TIME         = 40,           // resSource Creature, datalong - respawn delay.
+    SCRIPT_COMMAND_RESPAWN_DELAY            = 40,           // resSource Creature, datalong - respawn delay.
 };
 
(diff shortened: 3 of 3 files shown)


Ulduar:
Required for Patchwerk from the zone of death knights (final quest)
Also  required example:
Boss Anraphet(Cataclysm boss)
http://www.wowhead.com/npc=39788/anraphet
http://www.wowhead.com/npc=40251/stone-trogg-brute
http://www.wowhead.com/npc=39804/stone-trogg-pillager
http://www.wowhead.com/npc=40252/stone-trogg-rock-flinger
When 4 Elemental Guardian destroyed the room Anraphet, then himself appears to be the boss and down the stairs (after a short dialogue Brann), when it reaches the point of the movement (we will db_script_on_creature_movement), Anraphet says dialogue and casts a spell;
``` CPP
Id: 77437
Name: Протокол уничтожения
Description: Протокол уничтожения активирован. Все жизненные формы будут уничтожены.
ToolTip: Протокол уничтожения активирован. Все жизненные формы будут уничтожены.
General info
Category = 0, SpellIconId = 3175, ActiveIconId = 0, SpellVisual = (16352, 0)
SpellFamilyName = SPELLFAMILY_GENERAL, SpellFamilyFlags = 0x00000000 00000000 00000000
SpellSchoolMask = 32 (SCHOOL_MASK_SHADOW, SCHOOL_MASK_SPELL, SCHOOL_MASK_MAGIC, SCHOOL_MASK_ALL)
DamageClass = 0 (SPELL_DAMAGE_CLASS_NONE)
PreventionType = 0 (SPELL_PREVENTION_TYPE_NONE)
Attributes
AttributesEx1: 0x00000088 (SPELL_ATTR_EX_UNK3, SPELL_ATTR_EX_NEGATIVE)
AttributesEx2: 0x00000004 (SPELL_ATTR_EX2_CANT_REFLECTED)
Advanced info
Spell Level = 0, BaseLevel 0, MaxLevel 0, MaxTargetLevel 0
Category = 0
DispelType = 0 (DISPEL_NONE)
Mechanic = 0 (MECHANIC_NONE)
SpellRange (Id 13) "Anywhere"
MinRange = 0
MaxRange = 50000
Speed: 80.00
CastingTime (Id 1) = 0.00
Interrupt Flags: 0x00000000, AuraIF 0x00000000, ChannelIF 0x00000000
Chance = 0, charges - 0
Effects
Effect 0:
Id: 2 (SPELL_EFFECT_SCHOOL_DAMAGE)
BasePoints = 50
Targets (22, 15) (TARGET_CASTER_COORDINATES, TARGET_ALL_ENEMY_IN_AREA)
Radius (Id 22) 200.00
Effect 1:
Id: 3 (SPELL_EFFECT_DUMMY)
BasePoints = 0
Targets (22, 7) (TARGET_CASTER_COORDINATES, TARGET_AREAEFFECT_INSTANT)
Radius (Id 22) 200.00
Effect 2:
Id: 1 (SPELL_EFFECT_INSTAKILL)
BasePoints = 0
Targets (22, 7) (TARGET_CASTER_COORDINATES, TARGET_AREAEFFECT_INSTANT)
Radius (Id 22) 200.00
```
Effect 0 - dealdamage 50 Percent Hp for player.
Effect 1 - set respawn delay and corpse delay(for stone trogg and don't allow they revived)
Effect 2 - Instant kill all trogg(target script help) in room.
Good example, i think)
db_script_on_spell to help us effect 1)

xfurry:
@Grz3s do you think we require this for Patchwerk in DK area? I know you have been working on fixing the DB side of this.

Ulduar:
``` QUOTE
@Grz3s do you think we require this for Patchwerk in DK area? I know you have been working on fixing the DB side of this.
```
I'll tell you what does not. Respawn Time come with a sniff and a change in db - this is a hack.
And the need for more and Ronin (in dB change this hack). Plus Anraphet.

xfurry:
> I'll tell you what does not. Respawn Time come with a sniff and a change in db - this is a hack.
> And the need for more and Ronin (in db change this hack). Plus Anraphet.

I understand this.
Still, let's hear what other people have to say first. :wink:

Ulduar:
And why not just add? @Schmoozerd by the way was not against it!

xfurry:
Still, I think it doesn't hurt to hear the opinion of a fully qualified DB dev for this.
I'm not saying that your solution is bad, but I don't think that we have all the details of the problem yet.

So again, poke @Grz3s @Rog360 @virusav about the topic above.
We need to understand what is the exact issue with the Patchwerk and what is the best way to fix it.

Ulduar:
It is not just about Patchwerk, would you please read thoroughly. There are other examples.

Grz3s:
got nothing against adding new stuff..
about "Patchwerk" i gave my comment already here:
https://github.com/unified-db/Database/issues/13

Ulduar:
``` QUOTE
Then, I think it's safe to assume that this solution can be applied to all such similar cases.
```
@Xfurry Then there are other examples. Maybe for Patchwerk decision and hire, but not for others.
I gave an example of the Cataclysm, very dignified and meaningful.

xfurry:
We can investigate the cata issues.
But from the first look I think you would rather need a SCRIPT_COMMAND_RESPAWN rather than what you proposed.

Ulduar:
``` QUOTE
But from the first look I think you would rather need a SCRIPT_COMMAND_RESPAWN rather than what you proposed.
```
Nope, troggs in the Halls of Origination (Anraphet encounter) - change the respawn time, if you get them spell 77437, they do not respawn until you reset the instance, and if they kill another spell, then they quickly revived.
About Rhonin also do not forget to change the respawn time in db - full hack. Read again please from beginning to the end.

Ulduar:
Once again - Troggs by default have a respawn time of 10-15 seconds.
On Spellhit 77437 - Set Respawn Time of 43200(don't respawn untill you reset the instance)

xfurry:
Spell 77437 has a dummy effect.
You can just implement that effect in the core. In the end this is the purpose of the dummy spells, to have a server side script.

```
unitTarget->SetRespawnTime(999999.....);
```

For Rhonin I think a respawn command would work better, or maybe something similar with what we have for Patchwerk.


I said it before:
My concern is that allowing the change of respawn time or corpse delay time in DB might only open a door for DB issues. I don't think it's worth doing this just for one or a few unique cases.

Ulduar:
Are you kidding me? 
These cases, car and small truck, list all - life is not enough.
Ultraxion - should have the corpse delay 1 hour (config? Do not be ridiculous - it disappears after a minute, has no loot). His summon script and does not have a static spawn.
``` CPP
Spell 77437 has a dummy effect.
```
Db script than a worse?
``` QUOTE
DB might only open a door for DB issues.
```
Long overdue.

xfurry:
I'm not kidding you :smile: 

1. Ultraxion is not part of wotlk... I suppose this has to do with the cata core behavior or DB. So let's put him aside for now.
2. Let's stop using the word "hack" in stupid ways. The things that we are developing here are not hacks, and this has to be clear to everyone! Some solutions are better than others, but there are no such things as hacks.
3. For Rhonin the script needs to be fixed in some way. One easy way is just to teleport the boss back to his respawn location, instead of despawning.
4. As for your patch I still don't see the applicability of this. The corpse despawn delay can be simply handled by the SCRIPT_COMMAND_DESPAWN_SELF, while the respawn delay can be simply changed in DB if required, as grz3s showed you.
5. I can agree to a SCRIPT_COMMAND_SET_RESPAWN_DELAY, if this would be applicable to many cases. So far I can only see your example for Anraphet encounter, so I'd rather just put this in core for now. If you find more examples like this (and I mean good examples!) then we can have this dbscript command as you proposed it.

Ulduar:
Does anyone save me

Schmoozerd:
Just to be clear here:
We have mobs (which?) who have different respawn behaviour dependend on how they are killed?

Or is it just the case that some (which) mobs should respawn differently - but always after they decay?

Ulduar:
Explain. Patchwerk should not be delayed for the delay body. SCRIPT_COMMAND_DESPAWN_SELF - unwise. In the database, we can not change (in the config only)
Simply put Patchwerk died - Patchwerk corpse despawned immediately (at once, and not delayed)
Trogg in Hall of Origination(see link from first post)
When death respawn after  10 * IN_MILLISECONDS seconds.
Else if Spellhit 77437 - Set Respawn Time on 43200 * IN_MILLISECONDS.
To put it simply spell don't allow all trogg  respawn after death. Also do not have loot.
Ultraxion - Cata Bosses - https://github.com/Ulduar/scriptdev2-cata/blob/Dragon-Soul/scripts/kalimdor/caverns_of_time/dragon_soul/boss_ultraxion.cpp
Is Temporary Summon, has no loot (loot in the chest), but on the contrary it should be delayed for the disappearance of the corpse 1 Hour (offlike)
Set corpse in SD2 - bad idea, very bad.
Change is necessary not only to him, but still a more of NPC.
I think that these two commands in EventAI also be added.
https://github.com/ACID-Scripts/CATA/issues/1
All three bosses written Xfurry - thanks for the work. There is a problem.
EventAI script.
All 3 boss at EnterEvadeMode ();
Despawned and must respawn after 20 seconds. At the time of death varies respawn. At Blizzard as with all static spawn NPC.

Ulduar:
Addition about for Corpse Delay - see https://github.com/TrinityCore/TrinityCore/issues/4146
Analogic problem and more detail

Ulduar:
Close, hacks found path.

PR #150, opened 2015-04-01
Possibility to force enable/disable MMaps for specific creatures

fgreinus:
Not much to say here.

Adding the respecting new extraFlag to a creature's template results in disabling/enabling MMaps for this creature.
This may be useful for some raidbosses or some trashmobs (i.e. Al'ar).

Also we enabled MMaps for Players all the time and by default for pets too (but there it's possible to disable it by creature-template).

### DIFF
--- a/src/game/Creature.h
+++ b/src/game/Creature.h
@@ -43,19 +43,21 @@ struct GameEventCreatureData;
 
 enum CreatureFlagsExtra
 {
-    CREATURE_FLAG_EXTRA_INSTANCE_BIND   = 0x00000001,       // creature kill bind instance with killer and killer's group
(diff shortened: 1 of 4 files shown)


stfx:
Thanks for the thorough explanation. I think the PR is good but there are two little things which could be improved:
1. The enum flag gap is not necessary because those values are arbitrarily defined
2. Please only check for force enabled flag, player and pet if mmap is actually disabled using a boolean and force disabled flag if mmap is enabled. Like so:
```c++
bool isMMapForMapDisabled = g_mmapDisabledIds->find(mapId) == g_mmapDisabledIds->end();
if (isMMapForMapDisabled)
    if (unit->GetTypeId() == TYPEID_PLAYER)
    ...
```

fgreinus:
Okay, so i'll remove the flag gap.

Yes that do make sense to only do it, if it's disabled for map :+1:. I'll change this.

PR #151, opened 2015-05-05
Add stacking exception for Sanity Well and other spells, has iconid = 3062

Ulduar:


Ulduar:
I had to sacrifice the repository

PR #155, opened 2015-07-18
pet level modifiers for SPELL_EFFECT_SUMMON_PET

Tobschinski:
The multiplier field of SPELL_EFFECT_SUMMON_PET spell effects is added to
the master's level to yield the pet's level.

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -6648,7 +6648,10 @@ void Spell::EffectSummonPet(SpellEffectIndex eff_idx)
 
     NewSummon->SetRespawnCoord(pos);
 
-    uint32 petlevel = m_caster->getLevel();
+    uint32 petlevel = (m_caster->getLevel() + m_spellInfo->EffectMultipleValue[eff_idx] > 0)
+                      ? m_caster->getLevel() + m_spellInfo->EffectMultipleValue[eff_idx]
+                      : 1;
+
     NewSummon->setPetType(SUMMON_PET);
 
     uint32 faction = m_caster->getFaction();


Tobschinski:
Creatures summoned by this effect do not always have their master's level. The difference is given by the spell itself.

Example 1: 
Spell 6598 - Savannah Cub
BasePoints = 1, **Multiple = -13.00**
Master: http://www.wowhead.com/npc=3416/savannah-matriarch - **level 17 - 18**
Pet: http://www.wowhead.com/npc=5766/savannah-cub - **level 4 - 5**

Example 2:
Spell 12420 - Summon Skeletal Servant
BasePoints = 1, **Multiple = -3.00**
Master 1: http://www.wowhead.com/npc=10400/thuzadin-necromancer  - **level** 48 - **49**
Master 2: http://www.wowhead.com/npc=8553/necromancer  - **level 44** - 45
Pet: http://www.wowhead.com/npc=8477/skeletal-servant  - **level 41 - 46**

cala:
:+1: 
I poke @cyberium about this for his patch about the pet system : we discussed that SPELL_EFFECT_SUMMON_PET should use caster's level but now, we need to use also the modifier from @Tobschinski patch in order to have the correct level.

xfurry:
> If you want to avoid repeating code like this, here's an alternative:

@evil-at-wow 
```
Error	2	error C2782: 'const _Ty &std::max(const _Ty &,const _Ty &)' : template parameter '_Ty' is ambiguous	C:\...\Cmangos\workcore-wotlk\src\game\SpellEffects.cpp	6651

```

evil-at-wow:
The compiler doesn't know what type to use, which isn't really a surprise. So let's see what we have. The first part is adding the result of getLevel(), which returns a value of type uint32, with a value in EffectMultipleValue, which is a value of type float apparently. So we'll get an arithmetic conversion of uint32 to float, and the addition is done with floats. Which is a good thing here, because doing it as uint32 would yield an incorrect result (converting a negative float to uint32 is not a good idea). So, change the '1' to '1.0f' and you'll convince the compiler to use std::max( float, float ). The result then gets converted to uint32 and all is fine.

PR #156, opened 2015-08-25
Visual Studio 2015, ACE -> std+boost

namreeb:
See discussion here: http://cmangos.net/thread-7152.html


kvipka:
you have tested it? ) 
If it clearly Havoc version - it will not work

namreeb:
This is a polishing of Cyberiums work. I tested it locally and it seemed to work just fine.

kvipka:
test by login by 2 or more players in game in one time (2 or more windows of game). 
I remember, Cyberium has stoped work after this bug.

namreeb:
Yes, I did that. It seemed to work fine.

kvipka:
ok. Good. Then it's good compilation of commits in one =) Good job :)

cyberium:
Its nice now iam sure at least me and namreeb tested this.
But i need more tester. The process to merge it is simple. You have no reason to not test it.

Help us to make correct choice.

thank namreeb for this please add kvipka as contributor he helped me more than once.

Havoc:
For large scale server the current ACE implementation performs better I think. On Linux at least. Because on linux you will have to much overhead going on with Asio, because it emulates the async io of the proactor via a reactor implementation. More and more threads are getting involved. Also the locking overhead for small packet writes isn't solved yet. See Trinity's implementation for that issue. Trinity decides to workaround the problem by spawning special NetworkThreads for Linux which take care of writing the message queue to the socket (reactor style network io again). For Windows Server Asio will be good.

tarwyn:
Why is this considered for the master branch? If people want to experiment with alternatives to ACE, have them do it in a sub branch or in their own private repos. I don't care for it - I don't want to maintain Boost and I don't want to switch my build environment just to accomodate this.

namreeb:
This is not intended as an experiment.  It is intended to upgrade the project to a more modern version of the C++ language and to get rid of a poorly maintained and poorly documented and outdated dependency.  Note that the version of ACE that is being replaced will not even compile in a C++14 compliant compiler.

I can understand the desire not to replace one dependency for another, however I believe most will consider this a substantial upgrade.  It would be a shame for such upgrades to be delayed/prevented for lack of unanimous approval.

Zakamurite:
@namreeb I strongly believe that it can be compiled on any recent compiler, you must only deal with flags and defines.
Also the newest does not mean the best. It was already told that boost ASIO is much slower than ACE and I fully agree with it.

namreeb:
I will investigate the ASIO performance issue claim.

tarwyn:
@namreeb While I sympathize with your concerns to some degree, and appreciate the work you've put into this, I am pretty sure you can get Ace to compile on your compiler, without forcing everyone else to upgrade their build environments to match yours. 

Boost, the problem I have here is that it's a huge library that you're only using for a fraction of its features, and it's slower than ACE for the purpose it's used here. A lesser issue for me is the fact that it's not provided as part of the repo, thus adding overhead to maintain it (it's different obviously if you're already working with boost for something else).

Right now, we have a master branch that can be compiled all the way from VS 2009 - VS 2013. After this we would have what? 2013, 2015?

Looking through the open issues list, I find far more important candidates there that could use the effort over the effort required to get things to confirm to the latest C++ standard.

namreeb:
@tarwyn yes I'm quite sure I can get ACE to compile as well.  That is not the point.  The point is that I shouldn't have to be fiddling around with the headers of a third-party dependency. Another point is that I am skeptical about ACE performing better than boost in any case, unless our implementation is wrong.  That is what I will examine.  Also, can anyone tell me how many clients we're talking about before this alleged performance problem will manifest?  Are we talking about thousands of clients?  Tens?

boost should not and will not be included in the repository (at least by me).  It is prolific enough that it can reasonably be considered a standard component of a C++ build environment, in my opinion. 

As for what issues are more important, I humbly suggest that I have the freedom to work on those issues which I choose.

cyberium:
Today boost integration in CMaNGOS is matter of one merge to be accomplished.
All "known" big issues are resolved.
From Ambal to namreeb, all guys involved their time to make it at this point.
The boost lib prooved in many project that is good solution for any network project.
We have now opportunity to get a pretty recent lib, well supported and remove an old one. No need to compile or fix that new lib because not in the repos anymore and we just have to download it already compiled for out architecture. As simple as that.

ACE seem faster for some, its maybe right but:
I also eard ACE is more memory consuming and more buggy and more hack needed to make it work in majority of situation.
We have to update it.
Its always bad to have an external lib in a custom repo.
We have opportunity to reduce significantly the lines of code in CMaNGOS repos.

>Why is this considered for the master branch?

We talk and work on this since many years now, so its a bit late to say your surprise to see it considered for master.
>If people want to experiment with alternatives to ACE, have them do it in a sub branch or in their own private repos. I don't care for it - I don't want to maintain Boost and I don't want to switch my build environment just to accomodate this.

Then you'll understaind if we dont care you'll keep ACE in your own private repo.

tarwyn:
@Cyberium I understand you worked hard on it, but never merged it for it being not quite complete, and it was only picked up again on 08-18-2015, so excuse me if I do find this somewhat out-of-the-bluish. 

What exactly is broken about ACE that it needs removing? As far as I am concerned, it is working and working well. Please none of those "I heard" statements but facts please. And "My compiler doesn't compile it out of the box" doesn't count either.

In any case, considering the core is going nowhere in terms of getting the actual emulation feature-complete, I suppose my private repo is exactly what I'll be doing if this merge of your breaks my build environment.

namreeb:
Do you have any objections besides this alleged performance problem of ASIO in a UNIX environment?  From what I have read, such problems have long since been solved within ASIO.  As I said I would be happy to investigate this and if there is indeed a problem then I fully agree it should be resolved in some way before proceeding with this.  However a likely resolution may be to change how we use ASIO.

tarwyn:
@namreeb My biggest concern is for my build environment and the required C++11. I am not happy about Boost mostly for my own reasons, except for the fact that I am worrying that we're introducing a new POTENTIAL instability where right now there's none.

namreeb:
@tarwyn C++11 is already outdated with C++14 being the current standard.  This is why std:: types are used in this patch, rather than boost::, where possible.  Even Visual Studio has free versions of these modern compilers.  I see no reason not to expect that a modern compiler be used.

If someone wanted to write a new networking engine so that ASIO would not be needed, we could probably do without the other uses of boost as well, thereby eliminating the third party dependency altogether.  However, I am not myself convinced that this is a worthwhile use of time, so I do not plan to do it.

tarwyn:
@namreeb 
> If someone wanted to write a new networking engine so that ASIO would not be needed, we could probably do without the other uses of boost as well, thereby eliminating the third party dependency altogether. However, I am not myself convinced that this is a worthwhile use of time, so I do not plan to do it.

Agreed, that is time better spent on something else; no need to reinvent the wheel.

namreeb:
@tarwyn Right.  So I guess my point is that I feel like you're throwing out the baby with the bathwater.  My reasoning here is two-fold.  Firstly, even ACE is bigger than we need if we are accepting C++11 as a minimum standard (which is part of the purpose of this commit).  All we would be using from it, too, would be the networking component.  Secondly, boost is much more widely and actively used, reviewed and maintained.  I put boost in the same class as, for example, the openssl library.  It is common practice to require that a build environment be setup with openssl integrated into it.  This may be tedious initially, but ultimately shrinks the project size and also allows for future versions of the library to potentially be used without requiring our updating the project.

namreeb:
Note that I have amended the original commit to include a fix to the travis buildtest file.

cyberium:
What iam sure is, its very very hard to make everyone happy. But boost is definitivly a step forward.
Even if like @tarwyn said it may introduce some new instability at start.

Let now other team dev give their opinions.

Havoc:
What I wanted to say is that the server doesen't scale well with the current implementation as it is in this pull request. As reference see also: [Boost::Asio performance on UNIX-like OSes](https://github.com/TrinityCore/TrinityCore/issues/12932). This network library has too many abstraction layers and one more if you solve it like Trinity did. Of course it will work for small servers to play with some friends. We are talking about performance for thounds of players.

How would it be if we add the asio network library as standalone dependencie to the project from http://think-async.com. So we don't need boost. Most things could just be done with the standard C++11 library.

I also would recommend to drop all Visual Studio project files since we have Cmake as cross platform build solution. The generated Visual Studio project files are fine.

namreeb:
@Havoc I really like that idea, actually.  But is that not the same author as the primary contributor of boost.asio?  How do we know that the same performance problems will not exist here?

Havoc:
Yes its the same, but without boost. Now I think the performance doesen't matter here, since there is no interest in develop a server witch can handle thousands of players.

namreeb:
I see.  Yes, I like this idea.  However I wonder if we can solve both problems at the same time?

cyberium:
The main reason i stopped dev on boost is
- i knew that i will have no time to maintain it because of real life. Now this is changed.
- i knew some will not be happy so we have to take time to considere it with everyone involved in this project.
- i knew such a change will have some impact in how a lambda user will compile the project so a good documentation is needed.

So now as namreeb done a good job in perfect timing for me lets start the discution.

About Asio vs boost.asio i prefer boost.asio, because boost is large lib that we can use more than only asio in the future.
Also we already use at least boost::filesystem and boost::program_option.

Havoc:
Those boost dependencies could be rewritten. The question is do we need more features of the boost library in the future? If not, it would be more lightweight not using such a huge set of libraries.

tarwyn:
I apologize if I was a bit short-fused yesterday - hope nobody took any offense. Like @Havoc I prefer a lightweight option over the full monty Boost. With regards to using other features from Boost, I just ask myself is it worth the effort replacing existing systems that are NOT fundamentally broken? I guess that makes me the "if it ain't broke don't fix it" voice in the room. 
I also wonder if the lack of boost or strict C++11 coding standards stands in the way of us getting the Core feature complete, because if this is not the case, I feel we should focus on that goal first, and once we have stable, well working code, we can start polishing it. That way at least we have a baseline where we know that something worked at least once, which allows us to nail down issues stemming, for example, from adding Boost at a larger scale, much more effectively.

evil-at-wow:
Let me start by saying I do appreciate all the work and time that went into this. That said, this is one huge commit that "changes everything" to say so. A commit of the kind I wanted to avoid when I mentioned on the forum `let's not just blindly replace ACE by boost please`. My quote here is a bit too strong of course, because I'm sure you're not blindly replacing it. But as stated before, I would personally prefer to move in smaller steps. Even if we take the same steps and end up with the same result, this will make changes more manageable to review and discuss.

A few points and worries have already been brought up, so let me quickly add a few comments myself.

* Sure, ACE works. At least on the systems we've always supported and were it has been working fine, etc. If you have ever tried to build the core on let's say less common platforms, you've surely ran into some issues. Someone mentioned Debian has issues, and I've personally seen issues on my Linux system too. @Zakamurite has a valid point that you can most likely make it work (as I did on my system), but the fact remains that you can't just clone and build on some systems. I'm not saying that alone warrants replacing ACE, but it never hurt to see if there are any better alternatives.
* Asio vs boost.asio. Not much to say there. They are pretty similar except for namespace, so both could be used. Asio means less dependencies (and less code in the repository if we would choose to put it in the repository), but boost has some other libraries that could be useful. Call me neutral here.
* Asio speed. I'm sure we've all heard that ACE is faster, in particular on Linux. I haven't investigated this myself, so I don't have data to say otherwise, nor do I have reasons to believe that the claims are wrong.
* Build environment and C++. This is probably the most delicate change, because people usually don't like to change the way they work. I'm sure everyone understands we can't keep living in the past forever, so I do believe we need to make the step forward and start using (a part of) C++11. Now, there's a huge difference between taking a step forward, and running all the way to the other side. Perhaps @tarwyn was/is afraid we're running to the other side, and if that was his impression, I can't really blame him for that. I don't want to run all the way to the other side either.

So, let's see what we have and how we can take the next stop from here.
* It looks like the asio / boost::asio speed has to be investigated first.
* That means we can also postpone the asio vs boost::asio decision for now.
* We can start by introducing a subset of C++11, independent from the ACE vs asio discussion. For the time being, I wouldn't start dropping compilers yet for the sake of it and see where that can bring us (we can always decide to drop some if we have a good reason). For example, things like nullptr are in Visual Studio 2010 too. Let's make a list of what we'd like to use, and what compilers have the proper support.
* And last but not least, we can also start to minimize dependencies on ACE where they don't make sense (any more). Even if you like ACE as it is, you can't be against getting rid of ridiculous dependencies. For example, the vmap code also depends on ACE, and I believe there's no good reason for that any more. A bonus would be that some extractor tools are hopefully easier to build on the less common platforms! 

One question for @namreeb (and others, in fact). I've quickly scanned through the immense patch file, and I've noticed that you're using boost::uint32_t and so on instead of the standard C++ ones (see [http://en.cppreference.com/w/cpp/header/cstdint]). Is there any specific reason for this? Could we agree to use the standard C++ ones instead for the majority of the code?

I'll end my wall of text with repeating that I appreciate the effort you've all started here. I really do. I might not fully agree with everything you say or do, but that's OK. At least you've decided to step forward and do something to change things and get the ball rolling. And just that already deserves our respect :clap:

stfx:
I completely agree with @evil-at-wow on this one so make sure to reread his excellent reasoning again

namreeb:
I agree that we should investigate ASIO a bit.  I have sent an email to the asio-users@ email list inquiring about this issue.  I have also started writing a small application with the intent of simulating an environment which would demonstrate this issue.  If/when I finish it, I will publish it and any results I come up with.

As for switching to cstdint primitives, yes I agree we should do that.

namreeb:
I have now updated the commit replacing the boost:: primitives with std:: (including atomic types)

mityada:
I fully support ACE -> boost transition, but this pull request looks weird for me. It's a single huge commit which takes forever just to look through. Boost, C++11, CMake, VS and even code style fixes are mercilessly mixed together. Why not to rebase or cherry-pick all related commits from Ambal, Vladimir, Havoc, etc. repositories on top of current mangos tree? It would be much more easier to understand what have been done.

namreeb:
I was asked to squash them all into one.  Presumably to not clutter up the changelog, and to leave no chance that someone use some intermediary step, which may or may not compile.  While I would have been fine to do this with a series of commits, I am not motivated enough to go and re-do it now that it is done.

Schmoozerd:
A few more cents from me here (as absolute laymen when it comes to build environments)

* I seriously doubt such a fundamental (actually,  in its word) change can be reasonably split over multiple parts - sure: historically it was developped this way. But the start: ACE and end: Boost is clear and one changeset.

* Wishlist for modern C++: uint64 enums - such things will simplify the code in some places!

A few generic thoughts:
1) New is always better (except with whisky)
2) Politics: We cannot expect people to deliver such a workset multiple times - at some point, 1) must be uphold
3) Perfect is the opposite of good

A personal thought:
Personally: I do not care much about the build environment: It requires proper documentation, but if i need to update, i will update :)

mityada:
I think Asio performance issue may be mitigated by merging small packets into SMSG_COMPRESSED_MOVES (0x2FB). Despite the name, it could be used for any small packet, not only movement ones (at least on 1.12.1, but I suppose it should be same up to 3.3.5). I've found only two restrictions: the size of each packet could not exceed 255 bytes (including opcode) and SMSG_COMPRESSED_MOVES could be sent only when the player is in the world.

On retail, at least from MoP (unfortunately I don't have any sniffs from previous expansions), similar packet (SMSG_MULTIPLE_PACKETS) is intesively used by the server.

namreeb:
I am closing this as it is being re-done in favor of smaller commits.

namreeb:
As for SMSG_COMPRESSED_MOVES, there is another similar packet in the WotLK binary, called 'SMSG_COMPRESSED_UNKNOWN_1310' with value 0x1310.

The way the two packets are handled varies slightly.  See the following psuedo-code which handles both packets.  'isCompressedMoves' is true when the opcode is SMSG_COMPRESSED_MOVES.

```
void __stdcall ProcessNestedPacket(CDataStore *packet, int isCompressedMoves)
{
  CDataStore *v2; // eax@0
  CDataStore *this_; // esi@1
  void *v4; // esp@2
  unsigned int size_; // edi@2
  unsigned int read; // ebx@2
  unsigned int bytesRemaining; // ecx@2
  int v8; // eax@5
  int v9; // [sp+0h] [bp-4Ch]@2
  CDataStore currentPacket; // [sp+Ch] [bp-40h]@5
  CDataStore buffer; // [sp+24h] [bp-28h]@4
  unsigned __int8 *v12; // [sp+3Ch] [bp-10h]@5
  int size; // [sp+40h] [bp-Ch]@1
  int bytesRemaining_; // [sp+44h] [bp-8h]@1
  int a1; // [sp+48h] [bp-4h]@2

  this_ = v2;
  CDataStore::Get32(v2, &size);
  bytesRemaining_ = size;
  if ( isCompressedMoves )
  {
    v4 = alloca(size);
    size_ = this_->m_size;
    read = this_->m_read;
    bytesRemaining = this_->m_size - read;
    a1 = (int)&v9;
    CDataStore::GetDataInSitu(this_, (void **)&isCompressedMoves, bytesRemaining);
    sub_778180(a1, &bytesRemaining_, isCompressedMoves, size_ - read);
  }
  else
  {
    bytesRemaining_ = this_->m_size - this_->m_read;
    CDataStore::GetDataInSitu(this_, (void **)&a1, bytesRemaining_);
  }
  buffer.VMT = &gDataStoreVMT;
  buffer.m_buffer = (unsigned __int8 *)a1;
  buffer.m_base = 0;
  buffer.m_alloc = -1;
  buffer.m_size = bytesRemaining_;
  buffer.m_read = 0;
  if ( bytesRemaining_ )
  {
    do
    {
      CDataStore::Peak8(&buffer, (unsigned __int8 *)&isCompressedMoves + 3);
      CDataStore::GetDataInSitu(&buffer, (void **)&v12, BYTE3(isCompressedMoves));
      currentPacket.m_buffer = v12;
      currentPacket.VMT = &gDataStoreVMT;
      currentPacket.m_base = 0;
      currentPacket.m_alloc = -1;
      currentPacket.m_size = BYTE3(isCompressedMoves);
      currentPacket.m_read = 0;
      v8 = ClientServices::Get();
      sub_631FE0(v8, packet, &currentPacket, 0);
      currentPacket.VMT = &gDataStoreVMT;
      if ( currentPacket.m_alloc != -1 )
        gDataStoreVMT.InternalDestroy(
          &currentPacket,
          &currentPacket.m_buffer,
          &currentPacket.m_base,
          &currentPacket.m_alloc);
    }
    while ( buffer.m_read != buffer.m_size );
  }
  buffer.VMT = &gDataStoreVMT;
  if ( buffer.m_alloc != -1 )
    gDataStoreVMT.InternalDestroy(&buffer, &buffer.m_buffer, &buffer.m_base, &buffer.m_alloc);
}
```

mityada:
Decompression is done in sub_778180. So the only difference is that the SMSG_COMPRESSED_UNKNOWN_1310 is not compressed. Opcode name is a bit wrong :wink:

namreeb:
Yes, I edited my comment with updated psuedo-code.  In any case, this is not an ideal solution by itself.

PR #159, opened 2015-08-31
Implement hearbeat resist mechanic

Laizerox:
Spells with flag SPELL_ATTR_HEARTBEAT_RESIST_CHECK will have a chance to
break on each tick.

1) Creature will try to break aura based on time left 25, 50, 75% with a
base chance of 5% and 0,02% extra per resistance point into applied
school @ lvl 70 up to 10%

2) In PVP case resist checks are delayed by 5 seconds after which base %
will be calculated on amount of time remaining. In PVP case resistance school uses same formula as creatures and should apply resistance bonus given by talents/spells.

Note: Resistance bonus given by talents/spells doesn't appear to work properly if at all.
Note 2: Not tested in wotlk yet this was developed originally for classic and tbc and it's tested on it.


Laizerox:
Well this is not ready yet, but what about holy? each cc spell has it's own school ie, Druid root = nature and so on and resistance affects only them. I'm in process of rewriting this whole piece of code

cyberium:
@Laizerox any motivation to finish this?

Laizerox:
@Cyberium yes, I am actually refactoring a lot of this code.

cyberium:
Abandonned

PR #160, opened 2015-08-31
Use git_id for sql updates only

stfx:
Use git_id for sql updates only

This tool hinders workflow and is only useful for database updates. git_id will now silently exit except if new sql updates are found in which case it does the usual git_id sql magic.

Notes for devs:
* You need to rebuild git_id.exe
* Keep in mind that as before you need to push changes between git commits with sql updates as the last version is checked using the git remote.
* There is no longer any version check between script library (ScriptDev2) and core

Advantages:
* Reduce amount of change conflicts
* Able to merge PRs using github UI
* Less of a hassle when working in git with rebases, branches, etc.
* Make backporting easier and more fun

### DIFF
--- a/contrib/git_id/git_id.cpp
+++ b/contrib/git_id/git_id.cpp
@@ -50,16 +50,16 @@
 
 // config
 
-#define NUM_REMOTES 2
+#define NUM_REMOTES 3
 #define NUM_DATABASES 3
 
 char remotes[NUM_REMOTES][MAX_REMOTE] = {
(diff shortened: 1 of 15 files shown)


cyberium:
iam ok but let see what @xfurry think about this.

stfx:
Merged in https://github.com/cmangos/mangos-wotlk/commit/c77da8061677d6b6d224fa4c2ab0bb6fd4982e66

xfurry:
So much activity lately.... :hushed:

I need to catch up to everything before I can express my opinions.

PR #161, opened 2015-09-04
Various ACE to C++11 commits

namreeb:
Note we should add a commit dropping Visual Studio 2010 support as well as C++11 support via cmake for gcc/clang before this is merged.  I believe @evil-at-wow is working on at least the cmake part.


namreeb:
I'm closing this as much of it has been done separately since I opened this PR.  I will open a new PR with anything that may have been excluded.

PR #162, opened 2015-09-08
Update PvPstats table structure

ShinDarth:


xfurry:
Why are you doing this rename? It basically doesn't change anything

ShinDarth:
As I said in forum:

I don't remember exactly why, but all CMaNGOS cores are missing one update which renamed the field pvpstats_players.player_guid to pvpstats_players.character_guid (in the characters database).

That change is needed to fit with the web application. If I edit the web application I would break other things.

cyberium:
pull request seem fine but there is probably missing "alter db version"

the problem is i didnt followed the @stfx changes on git_id so iam not sure how to correct this PR.
It will be merged when i'll have time to look on it or when any other devs will want to merge it.

xfurry:
Also, please add ```[ci skip]``` to the commit message.

ShinDarth:
@xfurry done

stfx:
@cyberium git_id needs to be used for all commits with sql updates and thus for this as well

stfx:
@ShinDarth I see one more diff to TC table.
They have
`winner_faction tinyint(4) NOT NULL,`
while we have
`winner_team tinyint(3) NOT NULL,`
is this intended?

ShinDarth:
@stfx thank you to let me notice that. TrinityCore has its scheme updated with latest PvPstats web application revision, so we should have too:

```winner_faction tinyint(4) NOT NULL```

instead of 

```winner_team tinyint(3) NOT NULL,```

may you update it as well?

stfx:
@ShinDarth Feel free to provide a PR but please actually test it first because it is pretty clear that this hasn't been used by anyone in a while and we don't even know if it works.

PR #166, opened 2015-09-29
Fixed crash in instance resetting code in MapPersistentStateManager::_ResetOrWarnAll

namreeb:
DungeonPersistentState::UnbindThisState calls Player::UnbindInstance
Which calls DungeonPersistentState::RemovePlayer
Which calls MapPersistentState::UnloadIfEmpty
Which calls MapPersistentStateManager::RemovePersistentState.

MapPersistentStateManager::RemovePersistentState modifies the m_instanceSaveByInstanceId collection being iterated over in MapPersistentStateManager::_ResetOrWarnAll

Though perhaps it is cleaner to update the iterator upon removal, it would be a much more drastic change, and I think ultimately not as clean.

I am submitting this via pull request to give people an opportunity to review it.

### DIFF
--- a/src/game/MapPersistentStateMgr.cpp
+++ b/src/game/MapPersistentStateMgr.cpp
@@ -16,6 +16,8 @@
  * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
  */
 
+#include <list>
+
 #include "MapPersistentStateMgr.h"
 
 #include "SQLStorages.h"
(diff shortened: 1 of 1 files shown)


Kelzior:
Confirmed and working perfectly! No more crashes anymore when unbinding. Thanks

VladimirMangos:
Nice catch and fix!

namreeb:
I do not think a comment explaining this is necessary.  Everyone should know that you cannot modify a collection while iterating over it in this way.  However I will plan to add a brief explanation.

stfx:
I agree with @VladimirMangos

cyberium:
Yes nice catch i tough i was fixed that a year ago.

Kelzior:
I did report this some months ago but noone even replied then I contacted namreeb and he fixed it in 2mins

namreeb:
The comment has been added.  Merged.

PR #167, opened 2015-09-29
Removed TBB library.  

namreeb:
If anyone has any evidence as to why this is useful, we can re-visit it then.

Note that no objections were raised on the forum thread (located here: http://cmangos.net/thread-7192.html) for the past month.


VladimirMangos:
As I remember it mostly help at Unix side.... but can be wrong in memory.

namreeb:
It surely will reduce memory fragmentation, but so will some very basic pre-allocation.  It doesn't justify such a large third party API.  If there are some problems that appear as a result of this, we can consider other, simpler solutions.

sanctum32:
Core entirely doesn't use it at all... I didn't seen any calls of this library in core code. In pervious mangos experience I removed it and didn't seen any change in performance, memory usage.

namreeb:
It used the allocator, I believe.  But it did not use the custom containers.

zergtmn:
Iirc at the time it was added by @Ambal TBB allocator was much faster than standard allocator on Windows and slightly faster on linux. This might still be the case but I have no up-to-date information on this matter.

FallenAngelX:
https://github.com/cmangos/mangos-wotlk/blob/master/src/framework/Policies/MemoryManagement.cpp

namreeb:
I think there is no question that it is faster in almost call cases, sometimes more than twice as fast.  I'm just not sure that we need it?  It seems like an over optimization, and we will regain any overall performance lost once we implement per-map threading.

Ambal:
I added TBB library in order to utilize memory allocator + we wanted to reuse some threading primitives and thread-safe containers.

namreeb:
Closing this for now as it requires some benchmarking, analysis, etc..

PR #168, opened 2015-10-11
Fix travis clang build

DasBlub:
The clang build on travis has been broken (or rather: replaced by another
GCC build) since commit 7b68f38 because the install part always set GCC as
the compiler. Using built-in travis addons to install the sources fixes
this problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,9 @@ language: cpp
 git:
   depth: 1
 
+# make use of the new container based infrastructure at travis to improve performance (we don't need sudo)
+sudo: false
+
 # only run travis on the master branch
 branches:
   only:
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,14 +31,23 @@ compiler:
   - gcc
   - clang
 
+addons:
+  apt:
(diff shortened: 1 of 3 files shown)


DasBlub:
turns out the clang build has been broken after one of the resent commits (probably one of the C++11 related commits): https://travis-ci.org/cmangos/mangos-wotlk/jobs/84819418

```
/home/travis/build/cmangos/mangos-wotlk/dep/src/g3dlite/uint128.cpp:20:21: error: 

      non-constant-expression cannot be narrowed from type 'unsigned long' to

      'uint32' (aka 'unsigned int') in initializer list [-Wc++11-narrowing]

    uint32 a [2] = {_a & 0xFFFFFFFF, _a >> 32};

                    ^~~~~~~~~~~~~~~
```

cyberium:
seem like we have to either continue to hack g3d or update it.

DasBlub:
@cyberium can you look into the clang compile problems? i'm not familiar with either g3d nor the recent changes in the code base which now break the compile and i doubt that i'll have the time to dig into this.

PR #169, opened 2015-10-29
MC & Naxxramas - Missing Spells

Tobschinski:
MC - Firesworn - Thrash & Immolate
MC - Golemagg - Double Attack
MC - Core Rager - Thrash
Naxxramas - Anub'Rekhan - Double Attack
Naxxramas - Maexxna - Double Attack
Naxxramas - Gluth - Double Attack


Tobschinski:
Should be fine for all versions.

PR #172, opened 2015-11-05
Remove master only requirement

Rochet2:
This way forks can use travis without removing this and adding it back if a PR is made.
Not sure why this is here in the first place. If a branch should not be compiled then rather block that branch instead of blocking all branches.


cyberium:
Iam not against, maybe @DasBlub have another opinion?

Rochet2:
Hmm, maybe it could cause spam with this?
https://github.com/Rochet2/mangos-wotlk/blob/patch-2/.travis.yml#L57-L63

PR #173, opened 2015-11-09
Changed FindEquipSlot to call new func ViableEquipSlots

spkansas:
Split the look-up portion of FindEquipSlot into a second function ViableEquipSlots.  This is to be able to test or check for viability between two items, including those that are NOT in inventory.


Laizerox:
First of all I should point you to https://github.com/cmangos/issues/wiki/Coding-Standards because it may help to cleanup  your code and make bit better to read.

cyberium:
yes pleas redo this by using space instead of tab in your dev environement.

if you use visual studio -> https://msdn.microsoft.com/en-us/library/vstudio/ms165330%28v=vs.100%29.aspx
Select insert spaces, and verify the default is 4.

PR #175, opened 2015-11-10
Split FindEquipSlot code into ViableEquipSlots (Rev 3)

spkansas:
Split the look-up portion of FindEquipSlot into a second function ViableEquipSlots. This is to be able to test or check for viability between two items, including those that are NOT in inventory.  Fixed minor typo in debug code,


cyberium:
Pushed thank you ;)

PR #177, opened 2015-11-22
ACE to boost::asio.  thanks to shiro for the buffer design.

namreeb:
Note that this pull request will serve as a request for review.  It will remain open for 14 days before merging.


src/game/Player.cpp
```diff
@@ -2778,10 +2778,10 @@ void Player::SendInitialSpells()
 
     uint16 spellCount = 0;
 
-    WorldPacket data(SMSG_INITIAL_SPELLS, (1 + 2 + 4 * m_spells.size() + 2 + m_spellCooldowns.size() * (2 + 2 + 2 + 4 + 4)));
+    WorldPacket data(SMSG_INITIAL_SPELLS, (1 + 2 + 6 * m_spells.size() + 2 + m_spellCooldowns.size() * (2 + 2 + 2 + 4 + 4)));
```
tarwyn: Why the change here from 4 * to 6 * ?

src/game/WorldSession.cpp
```diff
@@ -102,17 +99,16 @@ WorldSession::~WorldSession()
...
-        m_Socket->RemoveReference();
-        m_Socket = nullptr;
-    }
+    if (!m_Socket->IsClosed())
+        m_Socket->Close();
 
     ///- empty incoming packet queue
-    WorldPacket* packet = nullptr;
-    while (_recvQueue.next(packet))
+    while (!m_packetQueue.empty())
```
dufernst: Nice to see this being changed in something that is a bit better supported.

It is a lot of code so I am just going to ask.
Are you sure nothing is being written to the queue at this point in time?

You close the socket before you cleanup this queue but the actual callbacks are triggered by a network thread.
Are you sure the network thread is not going to make calls to this object during its deletion/after it has been deleted?

Looking at your socket code I am fairly certain this albeit with a very small frequency might still result in either a crash and/or memory leaking.

Btw if you are making the switch to boost anyway, you could also consider using its lockfree queue to get rid of the read/write lock for this queue.

namreeb: Hello.  Thank you for taking the time to review this.
> Are you sure nothing is being written to the queue at this point in time?

Actually you are right.  A new lock is needed there.  I added a new commit for it here: https://github.com/namreeb/mangos-wotlk/commit/f2b43bfcd47071f4a40024ba0e0074979e94723a

We are making the switch to boost, but I believe we want to use C++11 types whenever possible.  That is why I did not use their container.

src/game/Player.cpp
namreeb: Because the arithmetic there was wrong.  It was likely up-ported from a previous version of CMaNGOS, maybe the TBC version.  The structure of this packet has changed.  Note that for each spell, a uint32 and a uint16 is written.  Therefore, the buffer will need six bytes per spell, not four.  This is not a major issue, but it was causing a re-allocation that was un-necessary.

src/game/WorldSession.cpp
dufernst: I don't think it solves the problem completely.

The network thread reads data en calls the function OnRead.
This function will then call ProcessIncomingData:
https://github.com/namreeb/mangos-wotlk/blob/boost.asio/src/game/WorldSocket.cpp#L140

In this function you check if the socket is already closed:
https://github.com/namreeb/mangos-wotlk/blob/boost.asio/src/game/WorldSocket.cpp#L197

However if you close it just after this check we end up calling QueuePacket (of the WorldSession):
https://github.com/namreeb/mangos-wotlk/blob/boost.asio/src/game/WorldSocket.cpp#L240

The world session might however already have been deleted resulting in a crash.

namreeb: Okay, I misunderstood you before.  I thought you were saying that there could be a double use of the queue.  I think that that was possible too, and that is what the commit I linked above should address.

If I understand you correctly now, the potential problem scenario is if World::UpdateSessions deletes the WorldSession at the same time that WorldSocket::ProcessIncomingData is executing, but only after the IsClosed() check.  Yes, that could be a problem.  Was it not a problem with the ACE networking code?  The only way I could see it not having been a problem before is if WorldSocket::ProcessIncomingData and World::UpdateSessions were guaranteed to be called from the same thread.

dufernst: You understood me correctly the first time as well, I asked multiple questions at the same time.
Possibly it might have been bugged before, the odds are very slim that it will cause a crash so it might have stayed undetected.

src/game/Player.cpp
tarwyn: Thanks for the explanation!

cala: I had the same question. Thanks for explaning! :+1:

src/shared/Network/Socket.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+void Socket::Close()
+{
+    assert(!IsClosed());
+
+    m_socket.shutdown(boost::asio::ip::tcp::socket::shutdown_both);
+    m_socket.close();
+
+    if (m_closeHandler)
+        m_closeHandler(this);
```
dufernst: The socket is here deleted, however the network thread might still execute callbacks of the socket resulting in crashes.
2 relatively easy ways so solve this are giving all callbacks a shared_ptr to the socket itself, in which case the socket cannot be deleted from memory until all callbacks are finished. Another very similar method involves giving all callbacks a weak_ptr to the socket, then you can try to get a shared_ptr to the socket in each callback.

namreeb: My understanding of the asio socket shutdown function is that it is supposed to immediately abort all read/write operations on the socket.  Is that not the case?

dufernst: It does indeed but the callbacks are still called by the network thread.
Therefore these could be handled before/after the object has been deleted.

Even if that is not the case the network thread might actually be already handling a callback function which could also result in a crash.

namreeb: Okay, I see your point.  Thank you again!  I will plan to look at this more closely as soon as I can, and will definitely try to resolve it before it gets merged.

cala:
Hi,
I tried to build this PR with CMake and gcc on a *NIX machine (Mac OS X) and it built without further adjustements needed, which is nice.
However, I can't start the core as I have an error mostly related to linking with boost:
```sh
$ ./realmd & ./mangosd 
[1] 22134
dyld: Library not loaded: libboodyld: Library not loaded: libboost_system.dylib
  Referenced frost_system.dylib
  Referenced from: /usr/local/mangos-wotlk/bin/.m: /usr/local/mangos-wotlk/bin/./realmd
  Reason: image not foun/mangosd
  Reason: image not foud
nd
[1]+  Trace/BPT trap          ./realmd
Trace/BPT trap
```
Note that I could build and start the core as of revision https://github.com/cmangos/mangos-wotlk/commit/feeadc7eaac1e38cf8ef5ef7ce67d5bb6fab6a61 which was after boost introduction.

namreeb:
Yes.  With this commit a dependency on the boost-system library was introduced.  Do you have the file, and cmake is for some reason not finding it?

cala:
I'll check this as soon as I have access to my build machine. I notice that this looks similar to the current travis build fail for this PR: https://travis-ci.org/cmangos/mangos-wotlk/jobs/92599981

However, the travis gcc build fails at the cmake step, not finding `boost_system`.
I made CMake work on my side using the `-DBOOST_INCLUDEDIR` option (the BOOST_ROOT variable seem to have no effect...) and made it point to the `include` directory of the boost build. The core build flawlessly, but can't start as shown above. To be precise, realmd does not start. So, I did not get to the starting mangosd step.

cyberium:
@namreeb 
For Travis, in travis.yml we have to add after packages needed lib.
like libboost-all-dev (wich is not a good idea)
better is to add libboost-system-dev. ( i am not sure we need only that one for now)

For includes iam not sure BOOST_ROOT is suited for linux. Is not linux have a standard place for all lib after installation? Well iam not sure at all.

For the socket problem any progress? Are you facing more issues?

namreeb:
@cyberium Okay, thanks.  I will look into that.  I may also adjust the cmake files to output the boost include and library paths explicitly.

As for the socket issue, I agree with @dufernst that using a std::shared_ptr<> for the sockets would solve the issues he has described.  In principle I don't like reference counted pointers, but this may be an acceptable use for them.  I have started switching to that and we'll see what it ends up looking like.

cala:
@namreeb I have the file:
```bash
$ ls -ls /usr/local/boost_1_59_0/lib/libboost_system.dylib
64 -rwxr-xr-x  1 root  wheel  30832 21 nov 15:42 /usr/local/boost_1_59_0/lib/libboost_system.dylib
```
Even  when running CMake with the explicit option `-DBOOST_LIBRARYDIR` the core (realmd + mangosd) doesn't start.

namreeb:
@cala Okay.  I will try and setup an OS X VM to see if I can troubleshoot it.

evil-at-wow:
@cala That's because the CMake option is only meant to set up a library search path for the linker, so the linker can find the boost libraries when it links the core binaries. But finding the boost libraries - or any libraries for that matter, which includes frameworks on Mac OS X - when you actually run the core is the task of the dynamic linker (dyld on Mac OS X), which is an operating system component. Of course, dyld doesn't have psychic powers, so the linker should help a bit by putting some info in the binaries, which no doubt happened but not entirely as expected.

If you run `otool -L realmd` you'll see that it lists all the shared libraries that the linker has set up for the realmd binary; and the expectation is of course that all these will be found by dyld at runtime. As you can imagine, using absolute paths for all libraries isn't exactly flexible and is typically only used for system libraries (and even then). So for some more flexibility, there are also 'variables' that dyld knows about and that the linker can use, to help with locating libraries: `@executable_path`, `@loader_path` (10.4+) and `@rpath` (10.5+). Don't worry about their exact meaning, I'm just telling you because you will see these getting used in otool's output.

Now, I don't know what we end up using in the core binaries these days, but if you could list the output you get from otool, that would give me a start.

In the meantime, you can help dyld to locate missing libraries on your system. Just set up a variable 'DYLD_LIBRARY_PATH' in your environment with a path to find the boost library. What path to use depends on what exactly was put your binaries by the linker (see otool), but assuming that CMake behaves a bit, you can probably run them by starting the core binaries with something like `DYLD_LIBRARY_PATH=/usr/local/boost_1_59_0/lib <realmd command>` or doing `export DYLD_LIBRARY_PATH=/usr/local/boost_1_59_0/lib` once before you type the command you normally type. Or, if you're lucky, you get a line like `@executable_path/libboost_system.dylib` or `@loader_path/libboost_system.dylib` in otool's output and you can get away with a simple symlink next to your binaries too (`ln -s /usr/local/boost_1_59_0/lib/libboost_system.dylib libboost_system.dylib` in the folder where your executables are).

cala:
Thanks guys for your help! :+1: 

@evil-at-wow I was also assuming that it was related with the linker/dynamic linker pair (which I know of) since dyld is mentioned in the error message. Thank you for confirming this to me. I did not know what the purpose of `otool` was until now. That is *really* helpful, thanks alot! (Someone less lazy than me would probably have run `man otool` long ago ;-) )

Here is its output for realmd and for mangosd respectively:
```sh
$ otool -L realmd 
realmd:
	/System/Library/Frameworks/Carbon.framework/Versions/A/Carbon (compatibility version 2.0.0, current version 152.0.0)
	/usr/local/mangos-wotlk/lib/libACE-5.8.3.dylib (compatibility version 0.0.0, current version 0.0.0)
	libboost_system.dylib (compatibility version 0.0.0, current version 0.0.0)
	libmysqlclient.18.dylib (compatibility version 18.0.0, current version 18.0.0)
	/usr/lib/libssl.0.9.8.dylib (compatibility version 0.9.8, current version 0.9.8)
	/usr/lib/libcrypto.0.9.8.dylib (compatibility version 0.9.8, current version 0.9.8)
	/usr/local/mangos-wotlk/lib/libtbb.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/local/mangos-wotlk/lib/libtbbmalloc.dylib (compatibility version 0.0.0, current version 0.0.0)
	/Users/mangos/toolchains/gcc-4.8.5/lib/libstdc++.6.dylib (compatibility version 7.0.0, current version 7.19.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 125.2.0)
	/Users/mangos/toolchains/gcc-4.8.5/lib/libgcc_s.1.dylib (compatibility version 1.0.0, current version 1.0.0)
```
```sh
$ otool -L mangosd 
mangosd:
	/System/Library/Frameworks/Carbon.framework/Versions/A/Carbon (compatibility version 2.0.0, current version 152.0.0)
	/usr/local/mangos-wotlk/lib/libACE-5.8.3.dylib (compatibility version 0.0.0, current version 0.0.0)
	libboost_system.dylib (compatibility version 0.0.0, current version 0.0.0)
	libmysqlclient.18.dylib (compatibility version 18.0.0, current version 18.0.0)
	/usr/lib/libssl.0.9.8.dylib (compatibility version 0.9.8, current version 0.9.8)
	/usr/lib/libcrypto.0.9.8.dylib (compatibility version 0.9.8, current version 0.9.8)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.3)
	/usr/local/mangos-wotlk/lib/libtbb.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/local/mangos-wotlk/lib/libtbbmalloc.dylib (compatibility version 0.0.0, current version 0.0.0)
	/Users/mangos/toolchains/gcc-4.8.5/lib/libstdc++.6.dylib (compatibility version 7.0.0, current version 7.19.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 125.2.0)
	/Users/mangos/toolchains/gcc-4.8.5/lib/libgcc_s.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/CoreServices.framework/Versions/A/CoreServices (compatibility version 1.0.0, current version 44.0.0)
```
As you can see, all are hard links with only two relative paths for mysql and boost. Yet, only boost lib raises an error, mysql lib is loaded. My `DYLD_LIBRARY_PATH` environment variable is currently empty. The only environment variable I'm aware of that points to mysql repository is `LIBPATH`:
```sh
$ echo $LIBPATH
/usr/local/lib:/usr/local/mysql/lib:
```
But adding boost lib directory to that variable and exporting it does nothing. However,  adding boost lib to `DYLD_LIBRARY_PATH` and exporting it do the trick: the lib is found and loaded (*i.e.* core starts).

But I'd like to now why this is not needed for mysql lib... Is this a setting on my side that I have to fix? A Mac OS setting that needs to be addressed either via specific installation steps or CMake option? A more generic issue in *NIX/CMake/gcc builds?

namreeb:
@dufernst, please see the additional commits I have added to the PR.

src/shared/Network/Socket.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+    return true;
+}
+
+void Socket::Close()
+{
+    assert(!IsClosed());
+
+    m_socket.shutdown(boost::asio::ip::tcp::socket::shutdown_both);
+    m_socket.close();
```
dufernst: Unless cmangos is supposed to run on exotic CPU's this is fine.

For CPU's without build in cache coherence (like I sadi: exotic) you have the following problem:
I don't think the close operation sets the closed state atomically, guaranteeing cache consistency for that variable.
You close the socket in the worldsession thread but read the closed state in the network thread.
They might see a different value, resulting again possibly in a crash.

dufernst:
The socket lifetime seems to be handled fine now.
I do however think you still might into a crash in the following scenario:

Here we check if the socket has been closed:
https://github.com/cmangos/mangos-wotlk/pull/177/files#diff-54d6b39b59a511bdd40ed529492dbcc5R89

And a few lines later we try to call ProcessIncomingData:
https://github.com/cmangos/mangos-wotlk/pull/177/files#diff-54d6b39b59a511bdd40ed529492dbcc5R110

Between the closed check and the call to ProcessIncomingData the worldsession might have been deleted resulting in a crash.
Some additional work is needed to manage the lifetime of the worldsession.
This can be solved with for example some more locking or by making the worldsession a shared_ptr and giving the socket a weak_ptr.

Btw: On blizzard you have some time to reconnect when your connection fails. Aka: the worldsession is not deleted as soon as the connection fails. If this would also be implemented (of course not in this PR) then the issue wouldn't have existed (unless the system cannot handle the load and the packet stream cannot be handled fast enough).

dufernst:
Actually you could also just store the messages in the socket object and then change:
https://github.com/cmangos/mangos-wotlk/pull/177/files#diff-1ee1ff863118a369830c9f2eefb5792fL201
to instead of reading from the queue in worldsession to let it read from the queue in the socket class instead.

In this case you don't need to change the lifetime management of the worldsession and it prevents access from the network thread to the worldsession altogether.

src/shared/Network/Socket.cpp
namreeb: Would that result in a crash?  I thought about adding an atomic boolean for 'closed', but it seemed like if there were an issue like the one you mentioned (admittedly I hadn't considered cache consistency across multiple cores in a CPU), I'm not sure that it would crash.  Worst case, the queued ASIO operations would be aborted, causing an error, which would cause a close, which would do nothing since by then all cores should have caught up to the initial close.

namreeb:
> Between the closed check and the call to ProcessIncomingData the worldsession might have been deleted resulting in a crash.

It is not clear to me why this would cause a crash.  Did you mean that the WorldSocket might have been deleted?  The WorldSession could not have been deleted without the socket being closed.  Therefore, the IsClosed() check in WorldSocket::ProcessIncomingData() would prevent any use of the deleted WorldSession instance.

As for the WorldSocket, it should only be finally deleted by the newly introduced cleanup thread once both the read and write states are marked as idle.  This will not happen in the read case until after the ProcessIncomingData() call is completed.

Incidentally, I realize that it is perhaps a bit overkill to create a new thread exclusively for cleaning up the sockets.  I plan to investigate whether it is possible to have the ASIO thread created by the NetworkThread class do this for us as a form of 'work'.

> Actually you could also just store the messages in the socket object and then change:
https://github.com/cmangos/mangos-wotlk/pull/177/files#diff-1ee1ff863118a369830c9f2eefb5792fL201
to instead of reading from the queue in worldsession to let it read from the queue in the socket class instead.

The problem with that is that it would involve introducing some new contextual awareness into the networking code.  The shared library would have to be able to perform the ARC4 crypto for the incoming and outgoing packet headers, track the ARC4 state, etc..  It would need that ability in order to parse the packet length reported from incoming client packets so as to divide up the buffer into packets.  I wanted to introducing that code into the shared library.

At some point in the hopefully-not-too-distant future, we may combine the framework and shared libraries into one single shared library that becomes its own repository, included as a submodule in the various versions of CMaNGOS.  This would also allow us move realmd into its own repository, as well.

src/shared/Network/Socket.cpp
dufernst: I guess it could result in a crash if because of that your IsClosed gives a false negative.
The IsClosed method just checks if the socket is closed, I don't know how boost handles that internally but if it just checks a non atomic variable the above scenario could happen.
https://github.com/cmangos/mangos-wotlk/pull/177/files#diff-8488fd1224d132d94707946bad856228R94

I doubt in any case that anyone is ever going to run it on CPU's that don't guarantee cache coherency.

dufernst:
> As for the WorldSocket, it should only be finally deleted by the newly introduced cleanup thread once both the read and write states are marked as idle. This will not happen in the read case until after the ProcessIncomingData() call is completed.

The lifetime of the WorldSocket is indeed fine.

> It is not clear to me why this would cause a crash. Did you mean that the WorldSocket might have been deleted? The WorldSession could not have been deleted without the socket being closed. Therefore, the IsClosed() check in WorldSocket::ProcessIncomingData() would prevent any use of the deleted WorldSession instance.

In ProcessIncomingData we check if the socket has been closed, this happens in the networking thread.
https://github.com/cmangos/mangos-wotlk/pull/177/files#diff-69deafea5e9d969f1439934ba282b2a7R195

In ProcessIncomingData QueuePacket is called on the WorldSession, this happens in the networking thread.
https://github.com/cmangos/mangos-wotlk/pull/177/files#diff-69deafea5e9d969f1439934ba282b2a7R238

It is possible that in between we call Close on the socket and finish deleting the WorldSession, this happens in the WorldSession thread.
https://github.com/cmangos/mangos-wotlk/pull/177/files#diff-1ee1ff863118a369830c9f2eefb5792fR103

evil-at-wow:
> As you can see, all are hard links with only two relative paths for mysql and boost. Yet, only boost lib raises an error, mysql lib is loaded. 

I have to say I expected the variables would be used for (most of) the non system libraries, but when I see your output I guess it's not a huge surprise to see them as full paths using the 'install prefix' (/usr/local/mangos-wotlk) or just two libaries. It might not be Mac OS X-like in the true sense, but then again, boost and MySQL probably just build their stuff on Mac OS X like they do on *nix in general and if that works, why not. And the reason why I say that it starts with the boost and MySQL build is because what you see here in the otool output is put in by the linker as such, but the linker gets that info from the actual libraries, so if you'd run `otool -D` on the boost library it would say 'libboost_system.dylib' too, and not something with a full path.

> My DYLD_LIBRARY_PATH environment variable is currently empty. The only environment variable I'm aware of that points to mysql repository is LIBPATH

That variable is not used by dyld and thus has no effect as far as I know, and [Apple's developer documentation](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/dyld.1.html) seems to confirm that. I can't actually test that now though to make 100% sure.

> But adding boost lib directory to that variable and exporting it does nothing.

That confirms what I just mentioned.

> However, adding boost lib to DYLD_LIBRARY_PATH and exporting it do the trick: the lib is found and loaded (i.e. core starts).

As expected, but it's always good to see expectations being confirmed in reality :smile: 

> But I'd like to now why this is not needed for mysql lib... Is this a setting on my side that I have to fix? A Mac OS setting that needs to be addressed either via specific installation steps or CMake option? A more generic issue in *NIX/CMake/gcc builds?

That's a good question, especially seeing you expect it to be in `/usr/local/mysql/lib`. Initially, when you reported that dyld did not find the boost library, I was not convinced yet that only the boost library was causing dyld to complain. I believed it just happened to be the first one, and dyld stopped after that because there's no point in spending more time doing something you already know won't work.

But if the core binaries actually load, there's only one possible conclusion really: dyld finds a suitable library. Either the one you expect it to find - although I don't see yet why it would find it - or another library somewhere on the system in a path that gets searched by dyld. I don't think that Apple has ever shipped MySQL libraries with the OS though, but perhaps you have ever installed one somehow (as part of another software install, homebrew, whatever). Let's find out which library it finds, for example by checking the opened files of the realmd process in Activity Monitor or by asking dyld itself (set/export DYLD_PRINT_LIBRARIES=1 too). Unless I see convincing evidence, I don't believe this is an error or setting that you have to fix.

namreeb:
@dufernst I apologize it's taken me so long to get back to you on this.  It is worth mentioning so we are both on the same page that with this patch, the WorldSocket owns the WorldSession.  I believe that it used to be the other way around.  This seemed like a reasonable change because you can have a WorldSocket without a WorldSession (specifically, before you enter the world with a specific character).

With this new ownership model in mind, I believe the best solution is to allow the WorldSocket to handle final destruction of the WorldSession.  This will have the side-effect of eliminating the problem you mentioned in your previous comment:

> It is possible that in between we call Close on the socket and finish deleting the WorldSession, this  happens in the WorldSession thread.
> https://github.com/cmangos/mangos-wotlk/pull/177/files#diff-1ee1ff863118a369830c9f2eefb5792fR103

> WorldSession is then already deleted resulting most likely in a crash.

I have begun to work on implementing this.

namreeb:
@dufernst It actually went quicker than I thought! :)

See the commit here: https://github.com/namreeb/mangos-wotlk/commit/6f1e5186520a96dacb788adef608cb8048b7d264

src/game/WorldSocket.cpp
```diff
@@ -258,496 +132,149 @@ int WorldSocket::open(void* a)
...
-                    // Its possible to cause deadlock if QueuePacket calls back
-                    m_Session->QueuePacket(new_pct);
-                    return 0;
+                    sLog.outError("WorldSocket::ProcessIncomingData: Client not authed opcode = %u", uint32(opcode));
+                    return false;
                 }
-                else
+
+                // if the player has logged out, don't bother queueing the packet.  instead, destroy the session >=]
+                if (m_session->PlayerLogout())
```
Shauren: WorldSession::PlayerLogout is only true DURING logout process - it is set back to false right after saving player to DB. You also need to keep in mind that session also includes the character selection screen (logout only means "get out of world"/delete player; )

dufernst: In general I don't think you can assume you will receive any more data from the client.
You now delete the session in ProcessIncomingData but this does not consider the case where the client server connection is suddenly aborted? Or would ProcessIncomingData still be called?

dufernst:
You close the socket in two places:
https://github.com/cmangos/mangos-wotlk/pull/177/files#diff-1ee1ff863118a369830c9f2eefb5792fR102
and
https://github.com/cmangos/mangos-wotlk/pull/177/files#diff-1ee1ff863118a369830c9f2eefb5792fR503

The second is used in normal logout scenario's and is fine.
The first should then take care of all remaining scenario's: e.g. connection abort.
As I already mentioned in my inline comment, are you sure that ProcessIncomingData is called when the connection is aborted? And if so will it always reach this section where the session is deleted now?

Other than that it looks thread safe, but as soon as it is cleared up if all the possible scenario's are handled by this implementation, I will take a better look at the code.
It is unfortunate that the code is getting hard to follow with all these modifications for thread safety.

src/game/WorldSocket.cpp
namreeb: Hmm, yes good point.  Okay, thank you.

namreeb: @Shauren @dufernst I believe commit https://github.com/namreeb/mangos-wotlk/commit/6494ff40690e245a8f2e9eaece86cd79d43eaa2b should resolve both issues.

src/game/WorldSocket.cpp
```diff
@@ -258,496 +131,141 @@ int WorldSocket::open(void* a)
...
-
     try
     {
         switch (opcode)
         {
-            case CMSG_PING:
-                return HandlePing(*new_pct);
             case CMSG_AUTH_SESSION:
-                if (m_Session)
+                if (!!m_session)
```
dufernst: Why the double ! here?
It behaves no different from `if (m_session)`.

src/mangosd/CMakeLists.txt
```diff
@@ -75,6 +75,7 @@ target_link_libraries(${EXECUTABLE_NAME}
   g3dlite
   gsoap
   ${ACE_LIBRARIES}
```
dufernst: ACE cannot be removed yet?

src/realmd/AuthSocket.cpp
```diff
@@ -198,51 +179,66 @@ AuthSocket::~AuthSocket()
...
-    BASIC_LOG("Accepting connection from '%s'", get_remote_address().c_str());
-}
-
 /// Read the packet from the client
-void AuthSocket::OnRead()
+bool AuthSocket::ProcessIncomingData()
 {
-    uint8 _cmd;
-    while (1)
+    // benchmarking has demonstrated that this lookup method is faster than std::map
```
dufernst: Just out of interest, is this lookup method also faster than std::unordered_map?
Not that it matters because for an array this small.

src/realmd/AuthSocket.cpp
```diff
@@ -1075,23 +1069,22 @@ bool AuthSocket::_HandleXferAccept()
...
-    patch_ = ACE_INVALID_HANDLE;
-
-    if (handler->open() == -1)
-    {
-        handler->close();
-        close_connection();
-    }
-}
-
+//void AuthSocket::InitPatch()
```
dufernst: Should this be deleted?

dufernst:
@namreeb in terms of multi threading everything looks good now.
If it also works as expected I think it can be merged.

src/game/WorldSocket.cpp
namreeb: It was to silence a compiler warning from Visual Studio and clang which complain about a performance hit in casting an integer type to a boolean.  I mistakenly thought that that applied to pointers, as well.  It will be removed from the final merge.

src/mangosd/CMakeLists.txt
namreeb: Not quite.  It's containers are still used in the database code, and it is also used I think in config file reading, and one or two other places.  We are very close once the ASIO implementation is merged, though.

src/realmd/AuthSocket.cpp
namreeb: I'm not sure, but I agree it doesn't really matter.  I suppose if we really wanted to micro-optimize this, we could change the order this array is defined with to list the presumptive most frequent opcodes first.

src/realmd/AuthSocket.cpp
namreeb: Yes.  It is worth mentioning that once the ASIO implementation is complete, I intend to pull realmd into its own repository.

namreeb:
Closing this since I will have to merge via git to resolve conflicts.

src/game/Player.cpp
cala: Poke myself to be careful about this when backporting.

namreeb: @cala If you're planning on backporting this to tbc and/or classic, I would suggest you wait.  We are close to removing ACE completely at which point I planned to backport this and my other recent commits myself.

cala: @namreeb Then I'll let you do it gladly: you will be far more efficient that I would ever be on such a matter. :smile:

PR #197, opened 2016-08-03
To remove the user, the database shold be selected first

dvdvideo1234:
mysql> use mysql;

Reading table information for completion of table and column names
You can turn off this feature to get a quicker startup with -A

Database changed
mysql> select user,host from user;

6 rows in set (0.00 sec)

mysql> DELETE FROM `mysql` . `user` WHERE CONVERT( USER USING utf8 ) = CONVERT( 'mangos' USING utf8 ) AND CONVERT( HOST USING utf8 ) = CONVERT( 'localhost' USING utf8 ) ;
Query OK, 1 row affected (0.06 sec)


dvdvideo1234:
The external database refresh is not linked to building the application

dvdvideo1234:
Whenever you guys prefer the flush or not I can always add it manually :dancer:

PR #201, opened 2016-10-14
Selectflags class & moving

Phatcat:
it's pretty obvious what they do; perhaps @xfurry or @killerwife wanna share their thoughts on these?

---
This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-wotlk/201)

### DIFF
--- a/src/game/Creature.cpp
+++ b/src/game/Creature.cpp
@@ -2286,6 +2286,12 @@ bool Creature::MeetsSelectAttackingRequirement(Unit* pTarget, SpellEntry const* 
         else if (selectFlags & SELECT_FLAG_NOT_IN_MELEE_RANGE && canReach)
             return false;
 
+        bool isMoving = pTarget->hasUnitState(UNIT_STAT_MOVING);
+        if (selectFlags & SELECT_FLAG_NOT_MOVING && isMoving)
+            return false;
+        else if (selectFlags & SELECT_FLAG_MOVING && !isMoving)
+            return false;
+
         if (selectFlags & SELECT_FLAG_IN_LOS && !IsWithinLOSInMap(pTarget))
             return false;
     }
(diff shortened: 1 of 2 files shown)


src/game/Creature.cpp
```diff
@@ -2247,6 +2247,27 @@ bool Creature::MeetsSelectAttackingRequirement(Unit* pTarget, SpellEntry const*
         if (selectFlags & SELECT_FLAG_PLAYER && pTarget->GetTypeId() != TYPEID_PLAYER)
             return false;
 
+        if (selectFlags & SELECT_FLAG_PLAYER_CLASS_WARRIOR && (pTarget->GetTypeId() != TYPEID_PLAYER || pTarget->getClass() != CLASS_WARRIOR))
```
Fabi: Are u serious with these if/else stuff? Rethink pls :D

killerwife: switch case for select flag, use some arbitrary select flag - number arithmetic, and make all class select flags use same block of code, inside one check for typeid player and in that IF so smth like if(select_flag-number!=getClass) return false;

Phatcat: switch case does not work in this case.

If we use a switch case, then you've moved the getclass to the beginning of the functionality, while it needs to be in the end, otherwise we may as well just have done a switch case on the power thing as well.

The reason why I do it as else if and not just a bunch of if's is because you are not able to send along 2 player selectoin flags, just as you are not able to send along 2 power selection flags, as with the current logic it would return false if it found the wrong class on the first, not even looking at the second flag - to solve this the flags would have to be reversed to the inverse, a NOT WARRIOR flag, and a NOT MANA flag, that way you're not excluding based on what people aren't, but rather on what they have.

I will repeat; this is the same method as used for finding power, if you have any thought out solutions please suggest them.

PR #204, opened 2016-11-10
Miscellaneous code cleanups.  Sorry for the large commit but there should be no functional changes here.

namreeb:
I am submitting this via PR to give people a chance to object.  Please note that this was exceptionally tedious, so please do not ask that it be broken up into small commits.  There should be no functional change as a result of this commit to surprise anyone.

### DIFF
--- a/src/framework/Utilities/Callback.h
+++ b/src/framework/Utilities/Callback.h
@@ -289,7 +289,7 @@ namespace MaNGOS
 
             typedef void (*Method)();
             Method m_method;
-            void _Execute() { (*m_method)(); }
+            void _Execute() const { (*m_method)(); }
 
         public:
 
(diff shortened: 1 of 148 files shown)


stfx:
👍 only missing whitespaces before static keyword changes and should we really use exceptions?

namreeb:
Why not use exceptions?

killerwife:
I think using exceptions in cmangos code is a divergence from code style used. We usually handle code without them. I think thats what @stfx was trying to say.

namreeb:
Okay well I can replace them with asserts since it's probably a topic for another discussion.  Incidentally, we do use exceptions already.  I don't see any reason why our style is benefited by discarding them.

killerwife:
Maybe in netcode/lower levels, but in general mechanics code, I have yet to see it.

cyberium:
👍  about throw exception its ok, its true that we used to use assert/MANGOS_ASSERT but it dont realy hurt to see that in code produced in 2016.

btw i wonder how you decided what method should be const or not there is for sure much case to handle.

namreeb:
I made const those functions which should not alter the state of their objects. I'm sure I missed some and there are some which could technically be made const but I left them since they do alter state.

PR #205, opened 2016-11-19
Replaced WorldSocket locking mechanism to prevent its destruction whi…

namreeb:
…le in use (use after free).  Also removed WorldSession references from WorldSocket.

### DIFF
--- a/src/game/WorldSession.cpp
+++ b/src/game/WorldSession.cpp
@@ -20,7 +20,7 @@
     \ingroup u2w
 */
 
-#include "WorldSocket.h"                                    // must be first to make ACE happy with ACE includes in it
+#include "WorldSocket.h"
 #include "Common.h"
 #include "Database/DatabaseEnv.h"
 #include "Log.h"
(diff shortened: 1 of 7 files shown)


namreeb:
@dufernst @Chaosvex @cyberium this is my (hopefully temporary) solution for the Socket use-after-free issue.  This should work until we transfer ownership of the socket away from NetworkThread.

Chaosvex:
I think `Increment` and `Decrement` would be more suitable names than `Lock` and `Unlock`, which imply some form of exclusive access.

namreeb:
If exclusivity is implied it is fine to rename, but Increment/Decrement alone is too vague I think.  How about AddReference() and RemoveReference()?

Chaosvex:
Anything along the lines of AddRef/RemoveRef/Release is fine in my book. I'm not a fan of needing explicit ref counting but that's another topic.

dufernst:
Even with the changes the code is not thread safe.

Lets assume the network thread just closed the connection and the session just entered the socket code but did not reach the Lock yet. In this scenario Deletable() returns true and the socket can be deleted.
https://github.com/cmangos/mangos-wotlk/pull/205/files#diff-8488fd1224d132d94707946bad856228R105

As soon as the session continues the server will crash.

PR #208, opened 2016-11-22
Enable MP-compilation for all projects

killerwife:
@namreeb


namreeb:
I think we agreed on IRC to postpone this for now, until VS2017 hits.

PR #209, opened 2016-11-26
Implement TriggerCastFlags

killerwife:


xfurry:
You should break this commit into many small parts. I will make it easier for us to understand.

killerwife:
@xfurry I already talked to cyberium a great deal about it and this is the smallest part it can be. It does 1 thing and 1 thing only: It replaces bool triggered with uint32 as CastSpell parameter so that we can start finding out which spell needs what. Its a huge change but it will allow us to swiftly add/remove TriggerCastFlags based on research done by people who script.

Main points: 
1. we misuse m_isTriggered = true in scripts
2. some spells need to ignore only part - @Grz3s found spell which only needs to target unselectable targets, I found that execute trigger needs to ignore hit calculation.
3. soft control over what to ignore and distinguish when it is actually a core bug and not need for triggered at all. I am quite sure that it will turn out that some spells have special properties.

The performance impact is close to zero, and all it will do is make triggers easier to find/change.

Writing this only as information for you. I already asked @Warlockbugs to merge it when he has time. Pushing this so quickly for 2 reasons: In the effort of exporting tbc-db I need all my tools at my disposal and @Grz3s needs it in wotlk.

Phatcat:
> Writing this only as information for you. I already asked @Warlockbugs to merge it

Don't forget we do peer reviewing for a reason. You are not excempt from peer reviewing because you've talked to cyberium or warlockbugs. This kind of cowboy merging gets on my nerves a bit.

> In the effort of exporting tbc-db I need all my tools at my disposal and @Grz3s needs it in wotlk.

Oh yeah? Well I need all my prs merged because of... reasons. I'll just go merge all of them because why not? Need me to tell you why this kind of merging is bad and why your reasons and invalid?

Also stop lying, you don't nreed this to export tbc-db, what a ridiculous proposition that is...

PR #212, opened 2016-12-05
Code export from VoA for Test Flight quests

killerwife:
Wotlk counterpart of 
https://github.com/cmangos/mangos-tbc/pull/149


cyberium:
i think that was implemented another way

killerwife:
It wasnt completely implemented. The part with CheckCast caster aura was skipped by you and needs to be finished generically.

PR #221, opened 2017-05-07
Fixed build g3dlite in X64 configuration.

Konctantin:
Fixed issue #Compilation error: identifier __cpuid not found #1367

Signed-off-by: Konstantin <gawrilyako@gmail.com>

### DIFF
--- a/dep/src/g3dlite/System.cpp
+++ b/dep/src/g3dlite/System.cpp
@@ -32,6 +32,7 @@
 
 #include <cstring>
 #include <cstdio>
+#include <intrin.h>
 
 // Uncomment the following line to turn off G3D::System memory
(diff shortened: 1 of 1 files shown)


Fabi:
This header file is for MSVC I think? 
What about updating the whole g3d lib instead (This is what I prefer, since it's solved there I think?!)
Or use an #ifdef for it

cyberium:
https://github.com/cmangos/mangos-classic/blob/1b3c8ee3c2dc90b0b0082bf2df640d48c9bcb96c/dep/src/g3dlite/System.cpp#L51
same solution but after the ifdef G3D_Win32
because it will fail in linux
So just wait next week that i'll port all of this.

PR #222, opened 2017-05-07
Fixed build g3dlite x64

Konctantin:


cyberium:
fixed another way thank you

PR #224, opened 2017-07-08
Extend EventAI to TotemAI

Phatcat:
Should allow for totems to have scripted behaviour through ACID

Which means all totems scripted through scriptdev can be moved to ACID, or at least most of them.

For instance;

https://github.com/cmangos/mangos-wotlk/blob/master/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/zulgurub/boss_jindo.cpp#L172
and
https://github.com/cmangos/mangos-wotlk/blob/master/src/game/AI/ScriptDevAI/scripts/northrend/borean_tundra.cpp#L1165

### DIFF
--- a/src/game/AI/BaseAI/TotemAI.cpp
+++ b/src/game/AI/BaseAI/TotemAI.cpp
@@ -33,7 +33,7 @@ int TotemAI::Permissible(const Creature* creature)
     return PERMIT_BASE_NO;
 }
 
-TotemAI::TotemAI(Creature* c) : CreatureAI(c)
+TotemAI::TotemAI(Creature* c) : CreatureEventAI(c)
 {
 }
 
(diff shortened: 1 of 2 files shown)


Grz3s:
Thank you ..for this patch.. 
Just tested and it works fine:
1st: Tested with EventAI
- .npc aiinfo - shows totemAI used
- npc behave normal - didn't notice anything suspicious (holds position, uses spells, aggro player when player aggro his master)
-  script added by EventAI - also works fine.

2. Tested Without EvenAI
- npc works like before patch (so all is ok)

Patch is reQ. sor some places like Magmoth (Borean Tundra) ... where totems not only support their masters .. but also do some visual Actions ( in this case they channel spells - in OOC)

:+1:

Grz3s:
@cyberium @killerwife any chance to look at this one?
Its the same thing... what we have with guardian_ai .. that we can use with eventAI together...
This patch is not harmful to anything.. and i tested it.. and did work well.
Thanks

cyberium:
Merged thank you both

PR #225, opened 2017-07-09
Add Areatrigger 171 into ScriptDev2

Grz3s:
•InstallFullDB.sh ... adds scriptdev2.sql after all DB updates, so
using DB update for scripted_areatriggers is not a good option. - Fixed
•Also - Cleanup for /* AREATRIGGER */ - we dont need to use
delete/insert if TRUNCATE scripted_areatrigger is used before.
•Cala;s update for UBRS implemented


xfurry:
@Grz3s if you decided to clean to top section of the file and remove the DELETE statements, then you should do the same for the other area triggers, that are added in specific instances.
Probably we should do the same for the scripted events table.

Grz3s:
should I move them into /* AREATRIGGER */ ..or leave them where they are?

PR #226, opened 2017-07-09
Add Areatrigger 171 into ScriptDev2

Grz3s:
•InstallFullDB.sh ... adds scriptdev2.sql after all DB updates, so
using DB update for scripted_areatriggers is not a good option. - Fixed
•Also - Cleanup for /* AREATRIGGER */ - we dont need to use
delete/insert if TRUNCATE scripted_areatrigger is used before.
•Cala;s update for UBRS implemented

### DIFF
--- a/sql/scriptdev2/scriptdev2.sql
+++ b/sql/scriptdev2/scriptdev2.sql
@@ -12,61 +12,8 @@ UPDATE world_template SET ScriptName='';
 /*  */
 
 /* AREATRIGGER */
-DELETE FROM scripted_areatrigger WHERE entry=4591;
-INSERT INTO scripted_areatrigger VALUES (4591,'at_coilfang_waterfall');
-DELETE FROM scripted_areatrigger WHERE entry=4560;
-INSERT INTO scripted_areatrigger VALUES (4560,'at_legion_teleporter');
-DELETE FROM scripted_areatrigger WHERE entry=3066;
-INSERT INTO scripted_areatrigger VALUES (3066,'at_ravenholdt');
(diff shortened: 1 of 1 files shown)


Grz3s:
@Xfurry wrote:
> @Grz3s if you decided to clean to top section of the file and remove the DELETE statements, then you should do the same for the other area triggers, that are added in specific instances.
 Probably we should do the same for the scripted events table.

- all scripted_areatriggers corrected
- scripted_event_id also

sql/scriptdev2/scriptdev2.sql
```diff
@@ -368,8 +334,6 @@ UPDATE instance_template SET ScriptName='instance_blackrock_spire' WHERE map=229
 UPDATE creature_template SET ScriptName='boss_overlord_wyrmthalak' WHERE entry=9568;
 UPDATE creature_template SET ScriptName='boss_gyth' WHERE entry=10339;
 UPDATE creature_template SET ScriptName='boss_pyroguard_emberseer' WHERE entry=9816;
-UPDATE creature_template SET ScriptName='npc_rookery_hatcher' WHERE entry=10683;
```
xfurry: Why delete the script for this creature?

Grz3s:
@xfurry fixed :+1: 

Also ..same thing is reQ for ClassicDB and TBC:
https://paste2.org/YIIz7hjJ -- areatrigger 171 added + CleanUp for Classic (includes: 9f9cb1f)
https://paste2.org/ZMyVdDaM -- areatrigger 171 added + CleanUp for TBCDB   (includes: b004cdd)

PR #227, opened 2017-07-10
[EventAI] update events; aura and missing aura

Phatcat:
There's no reason the npc needs to be in combat for this, do that logic
in the ACID script itself, don't hardcore enforce it here.

### DIFF
--- a/src/game/AI/EventAI/CreatureEventAI.cpp
+++ b/src/game/AI/EventAI/CreatureEventAI.cpp
@@ -403,9 +403,6 @@ bool CreatureEventAI::ProcessEvent(CreatureEventAIHolder& pHolder, Unit* pAction
             break;
         case EVENT_T_AURA:
         {
-            if (!m_creature->isInCombat())
-                return false;
-
             SpellAuraHolder* holder = m_creature->GetSpellAuraHolder(event.buffed.spellId);
             if (!holder || holder->GetStackAmount() < event.buffed.amount)
                 return false;
@@ -431,9 +428,6 @@ bool CreatureEventAI::ProcessEvent(CreatureEventAIHolder& pHolder, Unit* pAction
         }
         case EVENT_T_MISSING_AURA:
         {
-            if (!m_creature->isInCombat())
-                return false;
-
             SpellAuraHolder* holder = m_creature->GetSpellAuraHolder(event.buffed.spellId);
             if (holder && holder->GetStackAmount() >= event.buffed.amount)
                 return false;


Grz3s:
this is respond for this:
https://github.com/cmangos/issues/issues/1260

I tested ..and works perfectly fine:
As example we can use this creature:
```sql
Insert into `creature_ai_scripts` values
( '176901','1769','27','0','100','1','12544','1','15000','30000','11','12544','0','1','0','0','0','0','0','0','0','0','Moonrage Whitescalp - Cast Frost Armor on Missing Buff');
```

Also what is important here:
**Event - 23** is not used at all in ACID ... should I comment why? (koz it doesn't work like it should)
With this fix... finaly we can fix tons of casters ...that have some magic armors.
So instead of using OOC with timer every x secs... we can now use this event (aboved example is good here)

**Event - 27** is used many times in ACID. but gues what... :)
let me show one of them...
```sql
('231901','2319','1','0','100','1','1000','1000','1800000','1800000','11','12544','0','1','0','0','0','0','0','0','0','0','Syndicate Wizard - Cast Frost Armor on Spawn'), -- this trick works
(...)
('231909','2319','27','0','100','1','12544','1','15000','30000','11','12544','0','1','0','0','0','0','0','0','0','0','Syndicate Wizard - Cast Frost Armor on Missing Buff'), -- this one currently works in combat only --  and that's wrong
```
So As u can see from it... person who created it... knew that this event wont work ...so gave a timer on spawn (if u look correctly that's not even spawn event)

I strongly advice to add this update... and let us finally fix this old issue once for good.

Thanks to @Phatcat

killerwife:
While I think this provides only a temporary fix until creature group system is implemented, there is also zero harm in adding this PR.

Phatcat:
> While I think this provides only a temporary fix until creature group system is implemented

What... does this have to do with that? this is for self targeting...

killerwife:
Because its supposed to work for creature party members as well.

Phatcat:
... No... We have 16 = EVENT_T_FRIENDLY_MISSING_BUFF for checking if friendlies are missing buffs....

Phatcat:
>With this fix... finaly we can fix tons of casters ...that have some magic armors.

@killerwife these, casters with magic armors.. are those supposed to be cast on their party members as well? @Grz3s can this be true? are casters supposed to cast their magic armors on other friendly mobs? Otherwise event_t_aura and event_t_missing_aura is indeed still needed with these proposed changes for those kind of self checks alongside whatever needs to be added for wider party checking.

Grz3s:
no.. 
they will cast that "Magic" armor on themselves only.

Other type friendly spells can be done by Send/recive event

cyberium:
Merged thank you

PR #228, opened 2017-07-10
EventAI update

Phatcat:
~~2 new actions local evade and despawn guardians, 2 new targets, player invoker and player tapped, one updated action, quest event. check the individual commits for details.~~

Edit*; renamed local evade to reset combat as per suggestion and moved the 3 commits not related to resetting combat to another PR as per request.

Removed the option to not clear the threat list from my initial proposed commit; there is no reason you'd want to use any one of these 4 core functions that are called (removeauras, removethreatlist, cleartempfaction, stopcombat) without using the others, so making any one of these optional with parameters would be redundant, but there are a bucketload of quest mobs which potentially could very well be moved to db scripting / eventAI instead of being handled in the more heavy scriptdev using this.

There's someone like Balos Jacken in dustwallow (involved in the alliance vanilla quest The Deserters) as well as a good portion of the mobs from the alliance vanilla quest The Missing Diplomat (there's at least 3 mobs with this sort of behaviour just in that one questline, the dwarf in stormwind, the human in wetlands (which is the one used here linked by grz3s to showcase the use of all the 4 commits stemming from this pr) and the human in dustwallow). And that's 4 mobs just in the level 30 - 40 range.

### DIFF
--- a/doc/EventAI.txt
+++ b/doc/EventAI.txt
@@ -159,6 +159,7 @@ For all ACTION_T_RANDOM Actions, When a Particular Param is selected for the Eve
 50   ACTION_T_SET_REACT_STATE               ReactState                      Change react state of the creature
 51   ACTION_T_PAUSE_WAYPOINTS               DoPause                         Pause waypoints of creature
 52   ACTION_T_INTERRUPT_SPELL               SpellType - CurrentSpellTypes   Interrupt spell in given slot for creature
+53   ACTION_T_RESET_COMBAT                  No params                       Stops attacking, removes auras and clears threat list and clears temp. faction.
 
 * = Use -1 where the param is expected to do nothing. Random constant is generated for each event, so if you have a random yell and a random sound, they will be linked up with each other (ie. param2 with param2).
 
@@ -949,6 +950,12 @@ CURRENT_AUTOREPEAT_SPELL        = 2
 CURRENT_CHANNELED_SPELL         = 3
 Melee spells, generic spells and autorepeat spells have their interruptability based on interrupt flags, albeit this flag is wrongly missing for some spells.
 Main purpose of this command is to research and find channeled spells, which do not interrupt as a result of not having found an interrupt flag.
+
+---------------------------
+53 = ACTION_T_RESET_COMBAT:
+---------------------------
+This is used for mobs that need to stop attacking and leave combat without evading.
+Since the Threat List is cleared with combat stopping special targets are needed for events such as quest event after that point until resetting.
 
 =========================================
 Target Types
--- a/src/game/AI/EventAI/CreatureEventAI.cpp
+++ b/src/game/AI/EventAI/CreatureEventAI.cpp
@@ -1102,6 +1102,14 @@ void CreatureEventAI::ProcessAction(CreatureEventAI_Action const& action, uint32
             m_creature->InterruptSpell((CurrentSpellTypes)action.interruptSpell.currentSpellType);
             break;
         }
+        case ACTION_T_RESET_COMBAT:
+        {
+            m_creature->RemoveAllAurasOnEvade();
+            m_creature->ClearTemporaryFaction();
+            m_creature->CombatStop(true);
+            m_creature->DeleteThreatList();
+            break;
+        }
         default:
             sLog.outError("CreatureEventAi::ProcessAction(): action(%u) not implemented", static_cast<uint32>(action.type));
             break;
--- a/src/game/AI/EventAI/CreatureEventAI.h
+++ b/src/game/AI/EventAI/CreatureEventAI.h
@@ -129,6 +129,7 @@ enum EventAI_ActionType
     ACTION_T_SET_REACT_STATE            = 50,               // React state, unused, unused
     ACTION_T_PAUSE_WAYPOINTS            = 51,               // DoPause 0: unpause waypoint 1: pause waypoint, unused, unused
     ACTION_T_INTERRUPT_SPELL            = 52,               // SpellType enum CurrentSpellTypes, unused, unused
+    ACTION_T_RESET_COMBAT               = 53,               // No Params
 
     ACTION_T_END,
 };
--- a/src/game/AI/EventAI/CreatureEventAIMgr.cpp
+++ b/src/game/AI/EventAI/CreatureEventAIMgr.cpp
@@ -848,6 +848,7 @@ void CreatureEventAIMgr::LoadCreatureEventAI_Scripts()
                     case ACTION_T_FLEE_FOR_ASSIST:          // No Params
                     case ACTION_T_DIE:                      // No Params
                     case ACTION_T_ZONE_COMBAT_PULSE:        // No Params
+                    case ACTION_T_RESET_COMBAT:             // No Params
                     case ACTION_T_FORCE_DESPAWN:            // Delay
                     case ACTION_T_AUTO_ATTACK:              // AllowAttackState (0 = stop attack, anything else means continue attacking)
                     case ACTION_T_COMBAT_MOVEMENT:          // AllowCombatMovement (0 = stop combat based movement, anything else continue attacking)


killerwife:
"local evade" definitely not, it will only make a bigger mess of evade than it already is, someone has to first research and rework combat/evading to add such shenanigans
quest event - there is a core function that rewards it to whole group afaik, and it also needs to use constrains, what if player is across the continent?
despawn guardians - legit, +1
player invoker, player tapped - where is it used?

Phatcat:
> quest event - there is a core function that rewards it to whole group afaik, and it also needs to use constrains, what if player is across the continent?

It has constraints... it checks if they are in the same map... I can't use visibility because stealthed players would be counted out then

> "local evade" definitely not, it will only make a bigger mess of evade than it already is

It doesn't have anything to do with evade... it's not an evade function... it's a function which makes the unit stop fighting, clear threat list and remove temporary faction... It's needed for mobs which turn friendly during a fight. Which is where player invoker / player tapped comes in... If you wipe the threat list, then there's no target to award quest completion to. player tapped and player invoker is not cleared when combat stops, it's cleared on reset, therefore it will contain the player who first tapped the mob, and the player who stands to get kill credit.

> EDIT: ID 53 is reserved by me, I have an upcoming feature

.. Ehh.. Why not just use a later ID?

> there is a core function that rewards it to whole group afaik

there is a core function that runs AreaExploredOrEventHappens() on all members of a group? that could have saved me a bit of work, but great. And yes, it is needed to be able to award whole group, and no, action 26 is a mess and should be turned unused. No mobs use it in tbc-db and only one mob use it in wotlk-db that can be adjusted to this.

Grz3s:
"local evade" originaly was @xfurrys idea to help me with missing diplomat quest ... 
more here:
https://github.com/unified-db/Database/issues/74

Let me defend this and explain what it does.... because its something that we all need!
New Action 'Local Evade' (you can call it like u want) .... will basically... give us power with all quests like u know already Matis the Cruel...(or this missing diplomat.. or lots of others) to stop that fight in correct way.
- Creature will stop fighting with us
- Creature will drop thread list
- Player also wont continue fight
- AND whats most important here ...  he will stay where fight was finish !!! (this wond do any harm do normal evade)

I recorded avi with this patch,,, so u can look how it looks like...
https://youtu.be/GyLyES28WIk

Pls do not throw to bin something that we all need;]

killerwife:
And if used incorrectly, also wipe default movement system and breaking the creature on respawn. Not to mention what if you wanted to resume waypoints, or something similar. The reason why manipulating movegen is not done through DB commands is because its hard without tricky per case C++.

Phatcat:
> And if used incorrectly, also wipe default movement system and breaking the creature on respawn. Not to mention what if you wanted to resume waypoints, or something similar. The reason why manipulating movegen is not done through DB commands is because its hard without tricky per case C++.

Where am I manipulating movegen? I am removing temp faction, i am stopping combat and I am removing auras and clearing threat list. I am not manipulating the movement system, and doing acid wrong can break any creature.....

killerwife:
if MoveChase or any other combat gen is on the motion master stack, and you do your command, it will have unpredictable results if something pops the motion master stack.

Phatcat:
What command specifically?

RemoveAllAurasOnEvade(); ?

RemoveAllAurasOnEvade(); ?

DeleteThreatList(); ?

m_creature->CombatStop(true); ?

xfurry:
* The local evade might be still needed. The idea was to make a creature evade, without triggering the ```MoveTargetedHome()``` function, which would break the movement.
I don't remember exactly why I didn't go ahead and added this to the core. (maybe some issues which I didn't manage to fix).
* For the quest event update, I think there is already a function that does that. I think you need to use the function ```GroupEventHappens()``` which does exactly the same thing, but for the group (and also checks the distance).
* The new targets are nice to have. Same for the despawn guardian function. (albeit I'm not aware of any usage for these)

Phatcat:
> The local evade might be still needed.

Did you.... Watch the video grz3s linked to see how this functions? - also, I'll rename it from something else than evade if people can't work out that it DOES NOT PUT THE MOB IN EVADE MODE. I tried explaining this to killerwife who just tried making mountains out of molehills. (So where is the issue killerwife?, still waiting for your answer)

> For the quest event update, I think there is already a function that does that. I think you need to use the function GroupEventHappens() which does exactly the same thing, but for the group (and also checks the distance).

Yeah, I've updated to use that one now. Thank you. That one could have saved me a bit of time.

> The new targets are nice to have. Same for the despawn guardian function. (albeit I'm not aware of any usage for these)

So you didn't watch the video or read the sql he provided? everything I added here is used in that script (no, not both new targets are used, just one of them)

Phatcat:
I just honestly thought the name of action_t_stopcombatclearaurascleartempfactionclearthreatlist was a bad name. I see I was wrong and it is a much better name than local evade.

Warlockbugs:
ResetCombat is probably what you are looking at.

Phatcat:
> ResetCombat is probably what you are looking for, judging from description.

so you also just read the description and didn't check the video or how it works in effect. Thank you. Appreciate that.

Warlockbugs:
You are looking for a name for an action that acts just like evade, but does not actually moves npc to home location. Explain to me why ResetCombat does not fit?

I see it didn't took long before you go into passive aggressive mode all over again...

Phatcat:
> You are looking for a name for an action that acts just like evade

Actually I wasn't... Local_evade is a fine name... local_evade exactly has that in the name; it's en evade function on the spot... That people think that that must mean it involves EnterEvadeMode() I don't get why that is my problem; they could open the commit and look at th ecode and see it doesn't, if they truly cared.

> I see it didn't took long before you go into passive aggressive mode all over again...

You don't say what this

> ResetCombat is probably what you are looking for, judging from description.

 is in relation to, was I supposed to read that like xfurry said I was supposed to use that group event function, or am I to read this as a name suggestion? I am not a mind reader, Warlockbugs, no.

And was I wrong? did you even watch the video before dropping that comment?

Warlockbugs:
"Local evade" is extremely misleading name. Evade implies return to home location. We are looking for an element of evade, which is complete combat reset. So i suggest separating a common structural element for both, which would be ResetCombat.
The result would be 2 actions:
1) ResetCombat: Clears combat, resets auras, clears threat list, factions.
2) Evade: Calls ResetCombat, and then moves to home location.

killerwife:
Ok, I have thoroughly investigated the script that needs this and this was blown out of proportion entirely.

The script itself that started all of this is in SD2, its in wetlands and its buggy because its a complicated usage of EscortAI, DialogueHelper and other old features we usually do not use these days.

The reason why new targeting modes are needed, is so that we can reward credit to the person who engaged in combat with the AI. This is not entirely a great approach, the credit has to be awarded to the person who started the StartScript of the quest (i.e. clicked the accept on taking the quest). On a server where multiple people are competing for the completion, this might not work the way you want it to, i.e. people could grief each other. This is bad design #1. If the outlined issues are not clear, voice that opinion.

The whole combat stop thing. For starters, the evading, combat stopping and move home is underdeveloped. Evade is a specific term used by blizzard for when a creature has stopped combat and is briefly cannot be attacked due to running home. Hence local_evade is not an acceptable name in cmangos for this action. Now onto the functionality. It is essentially duplicated code of EnterEvadeMode with removed 2 calls and 1 call

killerwife:
Ok, I have thoroughly investigated the script that needs this and this was blown out of proportion entirely.

The script itself that started all of this is in SD2, its in wetlands and its buggy because its a complicated usage of EscortAI, DialogueHelper and other old features we usually do not use these days.

The reason why new targeting modes are needed, is so that we can reward credit to the person who engaged in combat with the AI. This is not entirely a great approach, the credit has to be awarded to the person who started the StartScript of the quest (i.e. clicked the accept on taking the quest). On a server where multiple people are competing for the completion, this might not work the way you want it to, i.e. people could grief each other. This is bad design #1. If the outlined issues are not clear, voice that opinion.

The whole combat stop thing. For starters, the evading, combat stopping and move home is underdeveloped. Evade is a specific term used by blizzard for when a creature has stopped combat and is briefly cannot be attacked due to running home. Hence local_evade is not an acceptable name in cmangos for this action. Now onto the functionality. It is essentially duplicated code of EnterEvadeMode with removed 2 calls and 1 call missing (triggering linked events). This means that if anytime in the future we change it, the EAI action will be broken. Also the action implementation does not properly deal with MotionMaster states, which could contain MoveChase and on respawn could potentially malfunction. As a result, this action is only acceptable if it is moved to a global AI concept available to all scripting mechanisms, not just EAI. Doing "just EAI" was a concept in the past and lead to numerous code duplicities in SD2, which means that maintaining many scripts is a time consuming task. I will think about the ramafications it can have and will deal with it once my big AI refactor is merged to cmangos. Expect it by the end of the month. Also note, I will most likely go with the suggested name "ResetCombat" because that's close to what that name means. Thx to @Warlockbugs for that suggestion.

Last but not least, Quest Event and Despawn guardians actions, these two are completely legitimate and I have 0 arguments against them, and can be merged effectively immediately.

Phatcat:
> the credit has to be awarded to the person who started the StartScript of the quest (i.e. clicked the accept on taking the quest).

Got any proof of this? Do you have any proof that shows that someone starting this quest, and not engagign in it, and someone else on the quest, engagign in it, who is not in the same group, that it should be the original script starter and not the one who stands to get kill credit. Show proof of this.

> On a server where multiple people are competing for the completion, this might not work the way you want it to, i.e. people could grief each other.

no, it is exactly working like I think it should, otherwise you could just basically force someone into doing the quest for you by camping at the questgiver and starting it over and over and over until someone kills him. THIS is bad design #1 killerwife....

> The whole combat stop thing. For starters, the evading, combat stopping and move home is underdeveloped. 

I'm gonna cry soon Killerwife.. this DOES NOT USE EVADING. IT DOES NOT USE MOVE HOME. It uses combat stop, but we use combat stop a lot of places in the core and in scripts. don't tell me it's suddenly a problem, now, here, specifically. If there is a problem, it's not with the usage. we should be able to tell a mob to stop combat specifically, so why should I not be able to here? 

> local_evade is not an acceptable name

Okay, fine, I'll call it something else.

> Now onto the functionality. It is essentially duplicated code of EnterEvadeMode with removed 2 calls and 1 call

Yeah but it's not, because it doesn't move the mob home, it is actually more duplicated code from scripts. Yes. this is a method that a lot of sd2 scripts use. To drop combat and remove the threat list. so sd2 scripts can do this but we can't do this with ACID even though if you would look at the video, you can see how well it works.

> This means that if anytime in the future we change it, the EAI action will be broken

Anything you change in the core acid uses can break ACID, will you stop with this nonsense already? it will also break all scripts using it then, and while all of those have to be fixed individually, acid could be fixed collectively, if it ever turned out to even be a problem in the future. you see? Mountains out of molehills... 'maybe possibly in the future things will work differently' ... Seriously...

>Also the action implementation does not properly deal with MotionMaster states, which could contain MoveChase and on respawn could potentially malfunction.

Seriously. I ask you for the last time now, killerwife. How?

I tell it to stop combat first, which means any combatmovement is WIPED. What are you talking about if it respawns perhaps it has combatmovement on the stack? from where killerwife?

> Doing "just EAI" was a concept in the past and lead to numerous code duplicities in SD2, which means that maintaining many scripts is a time consuming task.

Hey killerwife. You cannot tell me what to work on. I want to work on eventAI. Is that a problem? can't people work on eventAI`? - you can't demand that I expand what I work on other areas which it doesn't even affect!

killerwife:
Combat movement is wiped in MoveTargetedHome, not anywhere else. Or I should rather say all movegens above 1, not just combat movement. For the rest, well, you are asking me to provide research and you are the one making a PR. Also shouting, attacking me, or being passively aggressive will achieve only one thing, me not merging this PR. If that is what you wish to achieve, then continue.

Phatcat:
> you are asking me to provide research and you are the one making a PR

I provided a logical counter argument. Deal with that, then.

> no, it is exactly working like I think it should, otherwise you could just basically force someone into doing the quest for you by camping at the questgiver and starting it over and over and over until someone kills him. THIS is bad design #1 killerwife....

I'm not attacking you any more than you're attacking this. You're the one who've said the most, of all commenters, and provided the least solutions for any of the shortcomings or issues. you know that right?

> Combat movement is wiped in MoveTargetedHome, not anywhere else.

that means -any- creature which we depsawn or kill without evading, will not 'movetargethome', so all of those may have movechase when they respawn. Regardless of implementation of this or not. got any comment on that?

Phatcat:
> After that I suggest you look for other areas in wow-emu, be it vanilla/tbc/wotlk where a non-move home or a special combat stop with other means of moving is required and add them to the discussion.

There are lots of scripts that don't move home on evade..

https://github.com/cmangos/mangos-wotlk/blob/25f637a0427589daf6e06d56be8b4af012854510/src/game/AI/ScriptDevAI/scripts/kalimdor/ruins_of_ahnqiraj/boss_rajaxx.cpp#L169

https://github.com/cmangos/mangos-wotlk/blob/25f637a0427589daf6e06d56be8b4af012854510/src/game/AI/ScriptDevAI/scripts/outland/tempest_keep/arcatraz/boss_dalliah.cpp#L101

https://github.com/cmangos/mangos-wotlk/blob/25f637a0427589daf6e06d56be8b4af012854510/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/zulaman/boss_nalorakk.cpp#L134

https://github.com/cmangos/mangos-wotlk/blob/25f637a0427589daf6e06d56be8b4af012854510/src/game/AI/ScriptDevAI/scripts/northrend/crusaders_coliseum/trial_of_the_crusader/boss_jaraxxus.cpp#L131

Grz3s:
@Phatcat ... sory that by helping me ... you were put through all of this...

**Enough!**

cyberium:
I added an answer earlier in this morning but i probably not pressed comment button :/

To summarize it (and intentionally ignore for today aggressivity here)

ACTION_T_LOCAL_EVADE : The name is not suitable. Its not an evade!

A DB script is not a core script. It should follow different rules. Its not suitable for a DB script to do such a stuff on its own. It will require absolutely a following command or it will break that creature.
You cannot think all the case that may happen.
For this example its easy, the creature is de-spawned right after this command and should be reset at re-spawn. What about other kind of creature that may need this and will be only reset? I always defended new commands. But they should be coded to handle all possible issues. Is that creature can be possessed? What if pet are attacking? What if horde faction attack it? How it will react if player die or flee?

@Phatcat do not, never, compare an sd2 script with an db script, it will never be possible to put all sd2 power to DB without taking too much risk. Copy/Paste a part off an sd2 script and put it on a DB command is everything but a good idea. However you did a good job on this one its nice to see you proposing more stuff. Never forget anyway, you propose only then we discuss, then its merged or not. Also you have to admit that you have lack of knowledge of how core is working. So try to be more humble. Even more when i know that you made this for HELPING @Grz3s ;)
@Grz3s I know that you need more flexibility to implement some new script. I hope we'll found a suitable solution for this. Ah and all of this is only a discussion. With @Phatcat its always a bit animated but he is not too hard till now ;)

The @Warlockbugs solution is a good start (and avoid copy/paste by putting all code in same place) but only for core itself. (or sd2) This will not fix any issue with db script. As db script is limited, we need to wrap them with more check and fallback options. Enough bugged scripts.

For the rest i have same opinion that @xfurry 

@killerwife  let that for me or @xfurry

Phatcat:
> @Phatcat do not, never, compare an sd2 script with an db script

... I didn't... killerwife explicitly asked me for;

> After that I suggest you look for other areas in wow-emu, be it vanilla/tbc/wotlk where a non-move home or a special combat stop with other means of moving is required and add them to the discussion.

I'll repeat;

> **and add them to the discussion.**

If you look at my core PR for ZG cleanup you will see that I've moved mobs from acid to sd2 just to get more targeting capabilities. It's not like I'm trying to give ACID the full power of the core...

what is it that is so dangerous I am doing? Is is combatstop? is it dangerous to call combatstop? - Why? Because the mob will not 'move home' - so what? We have a ton of scripts, yes, sd2 scripts, who can do this fine, who can stop combat, clear threat list, all that stuff, NOT move home, what's that about then?

I would get it if your point was that this action is too specific, not general enough (which could be fixed with some of the params). but that's not the issue here. The issue is that it's 'unsafe' for some reason. I asked. How is this any more unsafe than despawning a mob which is currently in fight? That one will have chase movement as well and not be reset on respawn. Are all those mobs behaving 'unpredictably' or not functioning? or is it concievable that db devs can and have compensated for this?

> As db script is limited, we need to wrap them with more check and fallback options.

Wow this is amazing.. could this limitedness perhaps be because of... lack of functionality and flexibility in ACID framework? something I am trying to do something about? per request by a db dev... who you are vetoing saying a mixture of the mob being too complex for acid / my method is unsafe/unneeded.... because you clearly already made a decision;

> The @Warlockbugs solution is a good start (and avoid copy/paste by putting all code in same place) but only for core itself. (or sd2) This will not fix any issue with db script.

xfurry:
😱 OMG... walls of text on this thread. I need to book some time to be able to read through all of this.

cyberium:
>I didn't... killerwife explicitly asked me for;

You did... https://github.com/cmangos/mangos-wotlk/pull/228#issuecomment-314432473

> I'll repeat;...

No need to repeat, i think all of us here are able to read a comment.

> Wow this is amazing.. could this limitedness perhaps be because of... lack of functionality and flexibility in ACID framework?

Nice you start to understand. ACID is not designed to be able to do ALL the cases. And its nice to keep it simple but give it enough power to do a lot of stuff. Its also a priority to handle all border case that may cause creature/server going instable when we have to deal with DB command. DB data have by nature to be checked, managed, wrapped with failback solution. In not so long past adding an ACID was more hard than now. @Schmoozerd was way more conservative than us. And the history say he was right.

>  you clearly already made a decision.

Yes, you are smart enough to understand that code is not suitable as it is and i/we will not merge it if you will not work more on it.

Grz3s:
This topic has more comment that any other for long time now... its shows that still there is some life here:]

```quote
 Is that creature can be possessed? What if pet are attacking? What if horde faction attack it? How it will react if player die or flee?
```
Creatures during this..( from my experience) always... or change theirs faction to 35 or will get unitclass = notattackable... (or both)
So that means:
They cannot be attacked, possessed or whatever else ... all work in same way... surrender -> does smth -> despawn
So give me example ... to show me that i'm wrong here .. (because right now... all we hear .. is there is chance that something may happen ? but pls show me what:)

BTW: ... can evading creature be attacked, attacked by pet, possessed etc ??

Phatcat:
> Yes, you are smart enough to understand that code is not suitable

Actually. No. Cause you'll not give a reason -why-

I ask about how about despawned. Both you and killerwife go silent. How come.

I ask why it's not a problem sd2 scripts can do this and all you gotta say is 'don't compare sd2 with acidø' - that's not an argument, it's barely an answer.

> DB data have by nature to be checked, managed, wrapped with failback solution.

I love how you say this like sd2 scripts should not by nature be checked, managed and wrapped with fallback solutions. Oh right. They should. There's still no difference. 

but you are indeed 100% right. If you think I will pour any more energy into this here so you can keep telling me how little I know about the core (while having to explain to you all what player_invoker and player_tapped could even be used for, which, I don't think you guys have even gotten it yet) and how I am the one being hostile when the first comment I got was a comment telling me I can't use action 53 because it's reserved for some secret project? what has this place become, really..?

cyberium:
>Are you kidding me cyberium? Are you blind? you don't see killerwife explicitly asking me for that? Right there in the very same comment you link to. And then you're giving me a bs for it? for complying with something killerwife asked of me?
this is a joke, and if you are serious here, then this whole place is a joke.

Iam not a blind, it was easy for you to answer that sd2 is not comparable. But you choosed to link sd2 part of code. And from now i will certainly not do any joke with you.

> but you are indeed 100% right. If you think I will pour any more energy into this here so you can keep telling me how little I know about the core (while having to explain to you all what player_invoker and player_tapped could even be used for, which, I don't think you guys have even gotten it yet) and how I am the one being hostile when the first comment I got was a comment telling me I can't use action 53 because it's reserved for some secret project? what has this place become, really..?

I will not defend killerwife here, he is working on something that he will propose here and already added a command for it so he try to reduce amount of work on his side. Its bad i tell him. No one should know what is going on and its for the one who will merge the PR to rebase it.
Nothing secret. As you try to say... I already told you that i dont like how you are agressive. So this time i get enough patience. Banning from here since now. See you in another world.

cyberium:
@Grz3s this is not an evade.
I think you are better than me about testing a db script. So just try it and tell me the result. You have my 100% trust on this.
About that command especialy i repeat it may work on this case but its not a command that we can use easly without taking care more about the next step.
As we are why not add:

```
case ACTION_T_REMOVE_ALL_AURA_ON_EVADE:
  m_creature->RemoveAllAurasOnEvade();
  ...
case ACTION_T_LOCAL_CLEAR_TEMP_FACTION:
  m_creature->ClearTemporaryFaction();
  ...
case ACTION_T_LOCAL_COMBAT_STOP:
  m_creature->CombatStop(true);
  ...
```

You see the non sense?

xfurry:
Guys... Please stop.
Take a step back and relax. It literally took me 20 min to read through all the comments.

@Phatcat first thing: please split your PR into many small PRs.
Leave this topic with "local evade" separately.

@killerwife DialogueHelper class is still used btw. We are not creating any new content using that, because we kind of completed all the 3.x scripts that required it (maybe except ICC). Since there is no new content and we are not moving to 4.x or later, of course that DialogueHelper isn't used. Same applies for EscortAI or FollowerAI.

@cyberium and everyone else.
My original logic for this function, was to allow EventAI creatures to use the EnterEvadeMode function, without triggering MoveTargetedHome(). Basically this will allow us to script SD2 scripts (with escort or not) that require the creature to become friendly and evade, directly with DB script + eventAI.
Main difference, is that the DB script + eventAI will use the waypoint movement generator, and will be much more easy to write scripts. EventAI is required, because of the combat component.

Because of the complexity, such as our example: https://github.com/cmangos/mangos-wotlk/blob/master/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/wetlands.cpp#L134 we can't just trigger normal evade mode because that will trigger MoveTargetedHome(). And we don't want that, we want the creature to stay idle.
And, as you can see from the video posted by @Grz3s it works just fine.

We already use this kind of logic to script escort events in the Culling of Stratholme and Halls of Reflection. The only difference is that there we are handling the combat in SD2, while for our simple quest we want to do it in eventAI for simplicity reasons.
Links to the complex scripts below:
: https://github.com/cmangos/mangos-wotlk/blob/master/src/game/AI/ScriptDevAI/scripts/kalimdor/caverns_of_time/culling_of_stratholme/culling_of_stratholme.cpp
https://github.com/cmangos/mangos-wotlk/blob/master/src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/frozen_halls/halls_of_reflection/boss_lich_king.cpp 
https://raw.githubusercontent.com/unified-db/Database/master/Current_Release/Updates/408_updatepack_mangos.sql (search for Culling of stratholme)

cyberium:
@xfurry as you seem to know better what to do iam still waiting any action from you for this.

killerwife:
Combat should be moved out of unit class to support proper leashing, and add generic method enabling multiple movegens on evade. Until then, this should not be merged.

Ulduar:
@Phatcat 
My suggestions for your request
@Phatcat 
@cyberium 
Maybe add various variant.
``` CPP
 case ACTION_T_RESET_COMBAT:
        {
            switch (action.resetCombat.resetVariant)
            {
                case 0:
                  m_creature->RemoveAllAurasOnEvade();
                  m_creature->ClearTemporaryFaction();
                  m_creature->CombatStop(true);
                  m_creature->DeleteThreatList();
                  break;
                case 1:
                  m_creature->RemoveAllAurasOnDeath();
                  m_creature->ClearTemporaryFaction();
                  m_creature->CombatStop(true);
                  m_creature->DeleteThreatList();
                  break;
                case 2:
                  m_creature->CombatStop(true);
                  break;
                case 3:
                  m_creature->DeleteThreatList();
                  break;
                case 4:
                  m_creature->RemoveAllAurasOnDeath();
                  break;
                case 5:
                  m_creature->RemoveAllAurasOnEvade();
                  break;
                case 6:
                  m_creature->ClearTemporaryFaction();
                  break;
                case 7:
                  m_creature->AttackStop(true);
                  break;
            }
            break;
        }
```
``` CPP
// ACTION_T_RESET_COMBAT                         = 57
        struct
        {
            uint32 resetVariant;
            uint32 unused;
            uint32 unused1;
        } resetCombat;
```

xfurry:
> @xfurry as you seem to know better what to do iam still waiting any action from you for this.

Sorry, I forgot...
I was in holiday for the past weeks.
I will have a look again at this during the weekend.

PR #229, opened 2017-07-24
Playerbot in core

cala:
This PR adds @blueboy's playerbot (also known as portal) as an optional mod in CMaNGOS core.
By default, the core builds as usual, without any part of the mod code to prevent any stability or performance drawback. With the `BUILD_PLAYERBOT` cmake option, the mod will be built, embedded into the core.

I'd like people that don't use playerbot to check that they have no issues building, linking and installing the core as usual. The command `.bot add NameOfAnotherCharacterInSameAccount` should result in an error message: "no such command".

I'd also like people that intend to use playerbot to check that they have no issues building, linking and installing playerbot by adding `-DBUILD_PLAYERBOT=ON` to their cmake command. The command `.bot add NameOfAnotherCharacterInSameAccount` should then work as usual for them.

Tests for this PR should focus on building, linking and installing parts.

Original code by @blueboy and cie: https://github.com/blueboy/portal

This includes most of the updates made to the Playerbot API of Classic branch for the last 4 years. However, I did not include the changes made to each class AI due to the changes between Classic/TBC and WotLK: the original WotLK specific AI are kept.


madmax765:
Thank Cala, I sure hope this gets approved.  Love this player bot :D Thanks for keeping it up and going

PR #230, opened 2017-08-11
New cooldown system

cyberium:
The purpose of this is to unify all entities cooldown. With this change cd's are handled almost same way for all of them.
I also took that opportunity to fix some issues and add some features.
- now items will overwrite category spell at adding and at checking time.
- category cd can live without spell cd and vice versa.
- saving category and category time to keep correct track of the cd.
- removed the workaround to identify permanent cd.
- lockout can now be applied to creatures.
- added a command to show the list of active cd
- added a command to clear active cd only client side (can help for debug)
- added a command to clear active cd "arena way" (only cd < 10 min)

cd and gcd are now applied to creatures too. However it seem that they rarely use spell that have those cd. This have to be confirmed.

As always we'll all be happy of any comment about this.

### DIFF
--- a/src/game/Entities/PetHandler.cpp
+++ b/src/game/Entities/PetHandler.cpp
@@ -334,7 +334,7 @@ void WorldSession::HandlePetAction(WorldPacket& recv_data)
                         Spell::SendCastResult((Player*)owner, spellInfo, 0, result, true);
                 }
 
-                if (creature && !creature->IsSpellReady(*spellInfo))
+                if (creature && creature->IsSpellReady(*spellInfo))
                     GetPlayer()->SendClearCooldown(spellid, petUnit);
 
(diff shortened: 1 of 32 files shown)


src/game/Entities/PetHandler.cpp
```diff
@@ -337,7 +334,7 @@ void WorldSession::HandlePetAction(WorldPacket& recv_data)
                         Spell::SendCastResult((Player*)owner, spellInfo, 0, result, true);
                 }
 
-                if (creature && !creature->HasSpellCooldown(spellid))
+                if (creature && !creature->IsSpellReady(*spellInfo))
```
Konctantin: `!creature->IsSpellReady(*spellInfo)` mabe need `creature->IsSpellReady(*spellInfo)`

PR #231, opened 2017-08-12
Eventai update (2)

Phatcat:
Moved the 3 lesser controversial commits from the more heavily debated one per request, one has been merged already (thanks @killerwife), these are the remaining 2.

This commit; https://github.com/cmangos/mangos-wotlk/pull/231/commits/92229a1da183a27017092e1cdf91d3c936d420a5 makes this action; ACTION_T_QUEST_EVENT_ALL (26) redundant (deprecated) and the one single mob in wotlk-db which uses that action can with this commit use the party functionality from this commit in the normal ACTION_T_QUEST_EVENT (15) combined with one of the targets from the commit mentioned below and most likely be more blizzlike for it.

ACTION_T_CASTCREATUREGO_ALL (27) needs the same sort of attention. Rewarding creatureorgo to everyone on the threatlist will reward players who are not in party with the person/people who hold the kill credit. (same problem as with action 26) While this is not a problem in instances (obviously) - for world mobs it not only seems wrong, it is wrong, this sort of rewarding everyone with quest credit regardless if they hold the kill credit is something blizzard first introduced with Legion.

This commit: https://github.com/cmangos/mangos-wotlk/pull/231/commits/b4697b515b6b351964c3d56b74d6e766aaef8a8a allows the action ACTION_T_QUEST_EVENT (15) to select the correct target, the credit holder, rather than having to rely on the threat list which may be infected with a whole arsenal of people who do not hold the credit, especially top aggro may very well be a tanking class who is not in party with the person/group who pulled and holds the kill credit for the mob. Using TARGET_T_PLAYER_TAPPED in combination with the party functionality added in the commit mentioned above allows to correctly identify and reward the entire party that is supposed to be rewarded with the quest event according to kill credit. Using TARGET_T_PLAYER_INVOKER allows to correctly identify and reward the entire party that is supposed to be rewarded with the quest event according to who initially aggroed the mob.
This means that every mob which uses action 15 in eventai will probably have to be checked if their target can be optimised using one of those 2 new targets and if it also can be improved by rewarding the entire group using the added party functionality.

### DIFF
--- a/src/game/AI/EventAI/CreatureEventAI.cpp
+++ b/src/game/AI/EventAI/CreatureEventAI.cpp
@@ -767,14 +767,22 @@ void CreatureEventAI::ProcessAction(CreatureEventAI_Action const& action, uint32
             break;
         }
         case ACTION_T_QUEST_EVENT:
+        {
             if (Unit* target = GetTargetByType(action.quest_event.target, pActionInvoker, pAIEventSender, reportTargetError))
             {
                 if (target->GetTypeId() == TYPEID_PLAYER)
-                    ((Player*)target)->AreaExploredOrEventHappens(action.quest_event.questId);
+                {
+                    Player* playerTarget = static_cast<Player*>(target);
+                    if (action.quest_event.rewardGroup)
+                        playerTarget->GroupEventHappens(action.quest_event.questId, m_creature);
+                    else
+                        playerTarget->AreaExploredOrEventHappens(action.quest_event.questId);
+                }
             }
             else if (reportTargetError)
                 sLog.outErrorEventAI("Event %u - NULL target for ACTION_T_QUEST_EVENT(%u), target-type %u", EventId, action.type, action.quest_event.target);
             break;
+        }
         case ACTION_T_CAST_EVENT:
             if (Unit* target = GetTargetByType(action.cast_event.target, pActionInvoker, pAIEventSender, reportTargetError, 0, SELECT_FLAG_PLAYER))
             {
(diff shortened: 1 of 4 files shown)


src/game/AI/EventAI/CreatureEventAI.cpp
```diff
@@ -768,7 +768,12 @@ void CreatureEventAI::ProcessAction(CreatureEventAI_Action const& action, uint32
             if (Unit* target = GetTargetByType(action.quest_event.target, pActionInvoker, pAIEventSender, reportTargetError))
             {
                 if (target->GetTypeId() == TYPEID_PLAYER)
-                    ((Player*)target)->AreaExploredOrEventHappens(action.quest_event.questId);
+                {
+                    if (action.quest_event.rewardGroup)
+                        ((Player*)target)->GroupEventHappens(action.quest_event.questId, m_creature);
```
cyberium:
```cpp
if (Unit* target = GetTargetByType(action.quest_event.target, pActionInvoker, pAIEventSender, reportTargetError))
            {
                if (target->GetTypeId() == TYPEID_PLAYER)
                {
                    Player* playerTarget = static_cast<Player*>(target);
                    if (action.quest_event.rewardGroup)
                        playerTarget->GroupEventHappens(action.quest_event.questId, m_creature);
                    else
                        playerTarget->AreaExploredOrEventHappens(action.quest_event.questId);
                }
}
```

cyberium:
Merged Thank you

PR #236, opened 2017-09-08
Elrodan aurosalia dummy polymorph visual

Ulduar:
https://github.com/cmangos/wotlk-db/commit/1df0d35919a26b518768520776662b1307828e86
Maybe Send AI Event in random choice better, them Set Random Phase, if we have variant?
@xfurry, what do your think?


Ulduar:
Scripts not working right. Error compiling. I think, db variant more right.

PR #238, opened 2017-09-08
Confessor Paletress Reflectieve Shield

Ulduar:
http://www.wowhead.com/spell=66515/reflective-shield
Wraps the caster in a shield, absorbing damage and reflecting 25% of any absorbed damage back onto the attacker.
Remain damage / 4 (25 percent).

``` CPP
reflectSpell = 33619;
```
I am not sure about spell id.


cyberium:
Its not correct to use spell id exception in damage caclulation like that.

PR #239, opened 2017-09-11
Alar Credit quest.

Ulduar:
http://www.wowhead.com/quest=10946/ruse-of-the-ashtongue


Ulduar:
Close. Because error. Reopened in https://github.com/cmangos/mangos-wotlk/pull/240

PR #240, opened 2017-09-11
Alar kill credit, attempt two.

Ulduar2:
https://github.com/cmangos/mangos-wotlk/pull/239
I think, this is request not working (even not compiling, because so many error.)
Maybe this varian better?
@xfurry 
@X-Savior 
@boxa 
@evil-at-wow 
@cala 
@Tobschinski 
@Grz3s 
@cyberium 
please see, maybe need other repair?

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/outland/tempest_keep/the_eye/boss_alar.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/outland/tempest_keep/the_eye/boss_alar.cpp
@@ -108,6 +108,7 @@ struct boss_alarAI : public ScriptedAI
         // Start phase one and move to the closest platform
         m_uiPhase = PHASE_ONE;
         SetCombatMovement(false);
+        m_creature->SetWalk(false);
 
         m_uiRangeCheckTimer     = 0;
         m_uiCurrentPlatformId   = 0;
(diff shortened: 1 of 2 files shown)


killerwife:
1. Both commits should be in 1
2. Reuse the looting function, its called smth like IsEligibleForGroupRewards, instead of adding your own radius (I think this logic is unified, so when we change it there, its fixed elsewhere)
Apart from that, seems good.

Ulduar2:
``` Both commits should be in 1
```
On current state, i can use only with github editor.
``` IsEligibleForGroupRewards
```
I don't know about this method, sorry(

Ulduar2:
Pull redone 
https://github.com/cmangos/mangos-wotlk/pull/241

src/game/AI/ScriptDevAI/scripts/outland/tempest_keep/the_eye/boss_alar.cpp
```diff
@@ -144,6 +145,12 @@ struct boss_alarAI : public ScriptedAI
     {
         if (m_pInstance)
             m_pInstance->SetData(TYPE_ALAR, DONE);
+            
+        std::list<Player*> playerList;
+            GetPlayerListWithEntryInWorld(playerList, m_creature, 150.0f);
```
xfurry: I don't think you need to loop through the entire list of players.
The ones which are inside the instance, is enough.

Ulduar: Hello, Radu.
1) Pull redone. https://github.com/cmangos/mangos-wotlk/pull/241
2) I don't know, how make otherwise. I'm not such an advanced scripting writer(
For quest credit need quest and player should have aura for quest complete.

PR #241, opened 2017-09-11
Alar Credit Quest Condition.

Ulduar2:
If you have quest 10946 and has aura 42090 from quest item, after kill Alar - you will get quest credit.
Attempt number three.
@killerwife 
On current state, the only way I can combine two commits into one.

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/outland/tempest_keep/the_eye/boss_alar.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/outland/tempest_keep/the_eye/boss_alar.cpp
@@ -45,6 +45,9 @@ enum
     NPC_EMBER_OF_ALAR       = 19551,        // scripted in Acid
     NPC_FLAME_PATCH         = 20602,
     SPELL_FLAME_PATCH       = 35380,
+    
+    QUEST_RUSE_ASHTONGUE    = 10946,                // Quest 10946 for attunement in Black Temple.
+    SPELL_ASHTONGUE_RUSE    = 42090,                // Player can complete 10946 quest, only if has aura 42090. If kill Alar without this aura - quest not completed.
 
     MAX_PLATFORMS           = 4,
 
@@ -108,6 +111,7 @@ struct boss_alarAI : public ScriptedAI
         // Start phase one and move to the closest platform
         m_uiPhase = PHASE_ONE;
         SetCombatMovement(false);
+        m_creature->SetWalk(false);
 
         m_uiRangeCheckTimer     = 0;
         m_uiCurrentPlatformId   = 0;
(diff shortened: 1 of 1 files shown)


xfurry:
Well, I have a few comments about this PR.

1. You don't need to call the KillMonsterCredit function, because that is already happening when you kill the boss.
2. Actually, what we need to do, is to prevent the kill credit to be applied, when the player doesn't have the aura. Currently it's pretty tricky to do this.

killerwife:
Usually there is a hidden credit entry when that happens. Isnt there one for this q?

Ulduar:
``` CPP
You don't need to call the KillMonsterCredit function, because that is already happening when you kill the boss.
```
Need aura check for quest.
If don't need call the KillMonsterCredit function. What function need in this case?
``` CPP
Actually, what we need to do, is to prevent the kill credit to be applied, when the player doesn't have the aura. Currently it's pretty tricky to do this.
```
Because need script check. Alar kill and for credit need aura for player. If player not have aura - quest not completed.

``` CPP
Usually there is a hidden credit entry when that happens. Isnt there one for this q?
```
What method need use replace this?

killerwife:
Al'ar Kill Credit - 22850

xfurry:
> Usually there is a hidden credit entry when that happens. Isnt there one for this q?

@killerwife no. this is the same credit as the creature. no hidden entry here.



> Need aura check for quest.

@Ulduar  Yes, we do.



> Because need script check. Alar kill and for credit need aura for player. If player not have aura - quest not completed.

This is exactly the issue.
I don't think we can prevent the kill credit from happening, based on aura check.

The credit is being triggered here: https://github.com/cmangos/mangos-wotlk/blob/master/src/game/Entities/Unit.cpp#L849

xfurry:
> Al'ar Kill Credit - 22850

Quest 10946 requires kill credit for creature 19514 which is the boss itself. No kill credit entry.

killerwife:
quest_template data is completely guesswork, which means the data is simply wrong :) I can dig up sniff if you'd like.

xfurry:
> quest_template data is completely guesswork, which means the data is simply wrong :) I can dig up sniff if you'd like.

Would be great. If your guess is right, then it would be easy to fix.

killerwife:
Ok, tomorrow morning ill find it.

killerwife:
https://gist.github.com/killerwife/987f77ae6a3a942e6be4807bf9d0a5a9

So the data tells me its a "event" completed quest without any "kill credit".

xfurry:
Ok, then we need to update the quest_template data, and then we can implement the script

xfurry:
@killerwife any update on this? Do you have the correct quest template data?

killerwife:
@xfurry that was the data. It tells you that columns ReqCreatureOrGOId1-4 are 0 and that it should have QUEST_FLAGS_EVENT

xfurry:
Cool.
Well, I thought you will come up with a SQL query that will update the quest template data.
But now worries, I'll do that when I will push the complete patch.

killerwife:
I am doing 50 things at the same time, else I would've merged this myself. Sorry.

xfurry:
> I am doing 50 things at the same time

I can do it myself.
I would appreciate it if you can fix the passive flags issues.

killerwife:
https://github.com/cmangos/mangos-tbc/commits/feature/Relations - 66 commits and counting. Soon™.

PR #244, opened 2017-09-19
Add Halion remove phase aura on evade and death.

Ulduar:
Twilight Halion should remove aura on death too.


killerwife:
I am not all too familiar with Ruby Sanctum, but doesnt both 10-25normal/hc use same phase aura? Maybe it would be better to remove the particular phase aura and not all phase auras? That's the only thing I have with this PR.

Ulduar:
``` QUOTE
but doesnt both 10-25normal/hc use same phase aura? 
```
Now, use one aura for all mode.
``` QUOTE
Maybe it would be better to remove the particular phase aura and not all phase auras?
```

No, the Ruby Sanctum is not used for more than one phase of the aura

PR #245, opened 2017-09-22
Preevent die in air phase for Nightbane, Sapphiron and Felmyst.

Ulduar:


killerwife:
Damn. That's a nice one, though its surprising that none of my players managed to do this yet. Also possibly set HP to 1. Thats the customary blizz logic. Also I will add this to my list of "add to generic logic" because this is needed in plenty of scripts.

xfurry:
I would say that players should still be able to damage the boss. Maybe not kill it, but setting the damage to 0 doesn't look correct to me.

killerwife:
@xfurry the way it usually works is the mob is kept at 1 HP and then he dies when the invulnerability ends.

Ulduar:
@killerwife 
Generic solution can be add only in unit.cpp

For same case override in SD2 or EAI.

@xfurry 
Your suggestions?

killerwife:
Don't worry about it, I will add generic solution to AI :P This is good for merging imo.

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/boss_nightbane.cpp
```diff
@@ -82,12 +82,14 @@ struct boss_nightbaneAI : public npc_escortAI
     uint32 m_uiFireballBarrageTimer;
 
     bool m_bCombatStarted;
+    bool m_bNightbaneDamage;
```
killerwife: Why this variable? If you remove it, I can merge it.

Ulduar: It's done.

xfurry:
@killerwife I think the best solution here is to allow the boss to die at any time, but the corpse should fall down on the floor.

So basically the issue is that if the boss dies during the air phase, the players won't be able to reach the loot.

Laizerox:
@xfurry on retail it can't die in air phase

killerwife:
@xfurry what @Laizerox said is correct. Also, this is generic behaviour for many more bosses, so this should be implemented as some kind of base ScriptedAI capability for less code duplicity.

PR #248, opened 2017-09-25
Attempt add workaround for the Manticrone Cube.

Ulduar:
Sorry, i don't know, how realize in core side(


xfurry:
We can't accept workarounds.
I know that this event doesn't work and I already attempted a workaround a long time ago.
So we should better fix it properly this time.

PR #249, opened 2017-09-25
Attempt add new target TARGET_T_ACTION_INVOKER_ENEMY.

Ulduar:
@Schmoozerd when creating the Send AI Event system, mentioned a new type of target.
Sorry, i can't find link on issue(


cyberium:
@killerwife ;)

killerwife:
Ok, here is my verdict on this one. Since I reworked 70% of EAI internals, this is now unmergable. Next I don't have any examples of usage, which indicate why this is needed, and its very possible this is now possible without this event, since I remember adding some shenanigans. Also it uses old relations api, so it might behave differently from what was intended. @Ulduar should open an issue about what is not possible to do in EAI, and once we discuss that, we can do this PR in 10 minutes easy. Closing this for now.

PR #250, opened 2017-09-26
Add Berserk condition after 100 stacks SPELL Mark.

Ulduar:
Also avoid berserk spam.

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/northrend/naxxramas/boss_four_horsemen.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/northrend/naxxramas/boss_four_horsemen.cpp
@@ -88,6 +88,8 @@ enum
     SPELL_HOLY_BOLT         = 57376,
     SPELL_HOLY_BOLT_H       = 57465,
     SPELL_CONDEMNATION      = 57377,
+    
+    MAX_MARK_STACKS         = 100,
 
     // horseman spirits (not used in 3.x.x)
     // NPC_SPIRIT_OF_BLAUMEUX = 16776,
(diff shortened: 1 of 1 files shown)


xfurry:
I don't really understand what is this change supposed to do.
Can you please explain why do we need it, in more details.

Ulduar:
Hello @xfurry 
https://github.com/cmangos/mangos-wotlk/pull/250/files#diff-3aafe9716c7304e804e663f51a1ed42aR64
``` CPP
I don't really understand what is this change supposed to do.
Can you please explain why do we need it, in more details.
```

 // Note: Berserk should be applied once 100 marks are casted.

xfurry:
Ok, but is Berserk really cast after 100 stacks?
I thought that the stacks themselves server as the Berserk mechanics.

xfurry:
Poke @killerwife as you are the Naxx expert.
Should we cast Berserk after 100 stacks?

cala:
I'm not an expert of the WotLK version but I've done Naxxramas several times in Classic and TBC so yes, hitting the 100 marks was like the end of the encounter : in Classic/TBC the Horsemen would enrage and start using their abilities whitout cooldowns. In WotLK they would go berserk and have their damage significantly increased (x1000% o_O).

killerwife:
@xfurry I have no idea. Actually naxx is one of my weak spots, I never saw it in vanilla, and in wotlk it was so easy you didn't notice any of these mechanics.

cyberium:
Sound correct should we have to merge it? @xfurry maybe @Grz3s also.

cala:
I will deal with this (WotLK code and the Classic version).

cala:
The Classic version of this enrage is already in Classic & TBC core but needs some testing (I'm doing it atm). It is also done differently from this PR so I'll update either the PR or the two other cores for consistency.

cala:
Merged in https://github.com/cmangos/mangos-wotlk/commit/810bb8b2c6efc4caa55306dba1bee588e8bd8556 :+1:

PR #254, opened 2017-11-07
Add support script for spell 51854.

Ulduar2:
Add basic dummy event for Avatar of Freya.
Add preevent combat from NPC_FREYA_ACHIEV_TRIGGER.


xfurry:
Please provide more details on the purpose of this PR.
Which quest / event is this implementing and why do we need it in C++? Can't we do it with DBscripts only?

Ulduar2:
Happy New Year!
https://github.com/Furrazzlhf/RevivalsCore/blob/83197a879af54e9879bfc9cef001c793b4a90a81/src/server/scripts/Northrend/zone_sholazar_basin.cpp#L375
I'll find only this.
[https://www.youtube.com/watch?v=fqT_xaN23Bw](url)

PR #259, opened 2018-04-03
Fixed liquid processing in mmap builder

Shauren:
Also fixes detecting liquid id from LiquidType.dbc when entire grid is filling with the same type

This corrects my fuckup from 5 years ago.

Bonus fix (also related to liquids in mmaps), using switch() on flags is not a good idea.

### DIFF
--- a/contrib/extractor/System.cpp
+++ b/contrib/extractor/System.cpp
@@ -282,7 +282,7 @@ void ReadLiquidTypeTableDBC()
 
 // Map file format data
 static char const* MAP_MAGIC         = "MAPS";
-static char const* MAP_VERSION_MAGIC = "v1.3";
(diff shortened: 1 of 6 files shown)


killerwife:
:O @Shauren developing on wotlk :O

al3xc1985:
Shauren, you are really welcome here! We would like you to stay here forever :smile:

xfurry:
@killerwife is anyone looking at this? I can see you assigned yourself to the PR, but nothing happened.

killerwife:
Yes, I also investigated a lot of issues on TBC branch associated with it (which were not shaurens fault at all). I have a deadline IRL atm, so I do not have time to make all the usual exporting, but it will be done sometime in august.

PR #260, opened 2018-07-03
Add missing enum for Malygos script.

Ulduar:
Add Trinity Core copyright, because taken from
https://github.com/TrinityCore/TrinityCore/blob/master/src/server/scripts/Northrend/Nexus/EyeOfEternity/boss_malygos.cpp#L305
Malygos used method SetZoneOverrideLight for change light in combat.
Maybe need add opcodes for implement this method.

For Cmangos should be:
m_creature->GetMap()->SetZoneOverrideLight (zoneId, lightId, timer in millisecond);


xfurry:
This is interesting. I didn't know about this feature before.
We also need to implement the SMSG_OVERRIDE_LIGHT in order to have this. I would appreciate a full PR (or at least some more details).

Shauren:
This is referring to this https://github.com/TrinityCore/TrinityCore/commit/529cabea68ab90e34ccafa276fc40a8f2723902b

PR #262, opened 2018-10-12
Fix MM Hunter talent Wild Quiver

harlynx:
This PR aims to fix the currently non-functional MM Hunter talent 'Wild Quiver'. The spell data of the talent ranks (53215, 53216, 53217) is a bit inconsistent and posed quite a challenge to actually implement the aura proc. It utilises a currently unused aura that removes ammo cost from specific spells (which, as of 3.3.5 is only this talent).

The way the proc system in this specific case currently works is that it checks whether the mask of the proc causing spell matches the effect class mask, like a white list. Multishot, which matches the effect class mask and is a "Successful Ranged attack by Spell that uses a ranged weapon" triggers the proc even if it is not stated in the tooltip. I didn't play hunter during WotLK so I don't know if Wild Quiver worked that way, but the spell data implies it does so.

Spell 75 (Ranged Auto Attack) however does not appear in that list, so the [current proc handler implementation](https://github.com/cmangos/mangos-wotlk/blob/55b0b3b833911973d0a1f01bb3af3f32e48a2e5a/src/game/Spells/UnitAuraProcHandler.cpp#L606) rejects the aura proc because the mask check fails.
It appears to me that ProcFlag 06 (Successful Ranged attack (and wand spell cast)) seems to be exempt from the effect mask check.

I would appreciate some help on commit 9292ca1 because I think it feels out of place and maybe with your experience can be implemented better.

Tests when using this branch:

1. Learn any of the talent spells 53215 - 53217
2. .cast 53254 - it should not consume ammo
3. Auto attack or cast Multi Shot - it should sometimes fire an additional bullet (trigger 53254) without consuming ammo.

### DIFF
--- a/sql/updates/mangos/13983_01_mangos_spell_template.sql
+++ b/sql/updates/mangos/13983_01_mangos_spell_template.sql
@@ -1,0 +1,3 @@
+ALTER TABLE db_version CHANGE COLUMN required_13982_01_mangos_event_ai required_13983_01_mangos_spell_template bit;
+
+UPDATE `wotlkmangos`.`spell_template` SET `EffectSpellClassMask1_1` = 4097 WHERE id IN (53215, 53216, 53217);
--- a/src/game/Spells/Spell.cpp
+++ b/src/game/Spells/Spell.cpp
@@ -5082,7 +5082,9 @@ void Spell::TakeAmmo() const
 {
     if (m_attackType == RANGED_ATTACK && m_caster->GetTypeId() == TYPEID_PLAYER)
     {
-        Item* pItem = ((Player*)m_caster)->GetWeaponForAttack(RANGED_ATTACK, true, false);
+        auto player = (Player*)m_caster;
+
+        Item* pItem = player->GetWeaponForAttack(RANGED_ATTACK, true, false);
 
         // wands don't have ammo
         if (!pItem || pItem->GetProto()->SubClass == ITEM_SUBCLASS_WEAPON_WAND)
@@ -5093,17 +5095,17 @@ void Spell::TakeAmmo() const
             if (pItem->GetMaxStackCount() == 1)
             {
                 // decrease durability for non-stackable throw weapon
-                ((Player*)m_caster)->DurabilityPointLossForEquipSlot(EQUIPMENT_SLOT_RANGED);
+                player->DurabilityPointLossForEquipSlot(EQUIPMENT_SLOT_RANGED);
             }
             else
             {
                 // decrease items amount for stackable throw weapon
                 uint32 count = 1;
-                ((Player*)m_caster)->DestroyItemCount(pItem, count, true);
+                player->DestroyItemCount(pItem, count, true);
             }
         }
-        else if (uint32 ammo = ((Player*)m_caster)->GetUInt32Value(PLAYER_AMMO_ID))
-            ((Player*)m_caster)->DestroyItemCount(ammo, 1, true);
+        else if (uint32 ammo = player->GetUInt32Value(PLAYER_AMMO_ID))
+            player->DestroyItemCount(ammo, 1, true);
     }
 }
 
--- a/src/game/Spells/Spell.cpp
+++ b/src/game/Spells/Spell.cpp
@@ -5104,8 +5104,15 @@ void Spell::TakeAmmo() const
                 player->DestroyItemCount(pItem, count, true);
             }
         }
-        else if (uint32 ammo = player->GetUInt32Value(PLAYER_AMMO_ID))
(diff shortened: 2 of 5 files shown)


killerwife:
I have a few questions. Why are you including 6603 in the rule since the tooltip does not state that. Why not just add 0x0000000000000001 mask (auto shot mask spellid 75) to the proc trigger mask, thereby including auto-shot for the wild quiver.

Also I would squash the whole thing and name the commit "Implement SPELL_AURA_ABILITY_CONSUME_NO_AMMO" since thats what it essentially does.

harlynx:
I included the melee attack spell for it to be consistent. It's definitely out of the scope of this fix though.

If you mean adding the mask to the mysql  database: that's what I was considering as well but I didn't want to alter Blizzard's spell data because I didn't know if this was common practice. If that's the case I am happy to discard the last commit.

killerwife:
I absolutely forgot to reply to you yday. Autoshot and melee attack are very different, hence the consistency argument is out of the question, unless you find mentions of it from live realms that wild quiver worked on melee (which I doubt it did since I played hunter, but my memory is hazy).

Next, the masks for spells are for example completely nonexistant in vanilla and tbc due to obfuscation, and when present even wrong. I doubt this particular mask had an impact in client, hence it might as well be faulty data, and there is no way to know for sure. Try to add the autoshot mask to effect1 and if it works, then might as well be done with it.

harlynx:
Yes I did not intend for it to work with melee attacks (it wouldn't due to the proc flags) but rather in general if there are other edge cases - but that's probably not a good idea. Who knows what spells will start to behave differently.

This is all very fascinating to me though, also the thing with the obfuscation. Thanks for the review. You can squash them all into one commit.

killerwife:
Note: we dont do DB revisions with spell.sql and also never make a revision yourself in a PR, the merger does that. In future you would do 00000_* etc.

harlynx:
Okay, thanks. I will keep it in mind for next time.

src/game/Spells/Spell.cpp
```diff
@@ -5082,7 +5082,9 @@ void Spell::TakeAmmo() const
 {
     if (m_attackType == RANGED_ATTACK && m_caster->GetTypeId() == TYPEID_PLAYER)
     {
-        Item* pItem = ((Player*)m_caster)->GetWeaponForAttack(RANGED_ATTACK, true, false);
+        auto player = (Player*)m_caster;
```
P-Kito: Why would you use `auto` for a trivial datatype? 🤐

cyberium: Why not? Using it on non trivial type is better? Reading that line you know directly what type will be assigned. Devs are lazy and there is 3 keypress less in "auto" than "Player*". Almost 50% less typing... Iam not so against :p

killerwife: I guess it just comes down to me and how I work with pointers, references and how I keep them in my head. I know that many people have issues with keeping track of it(C# does not have this).  If its 1.5-2.x bigger its a non issue, when its something like std::vector<std::map<>>.... the tool to use is typedef instead, again, at least for me. The reality is that for a dev which is 100% dev in his time, this is non-issue, but most people contributing are not, hence its always great when they have to write that asterisk.

Also now that I think about it, the bigger thing here is that C-Style cast, should be static_cast instead.

harlynx: I used auto because I am used to it. At work we have modern code bases (C++17) and there we mostly stick to the IsoCpp core guidelines which encourage auto in most cases - there are exceptions but usually you can trust the compiler's ability to infer types. I know this code base isn't 'modern', and the issue was not about refactoring a function, but adding functionality instead, which is why I left it that way. To me it is apparent that a variable called `player` is an instance of `Player`. I definitely did not want to add yet another line of `(Player*)m_caster)->` to that function. 😅

sql/updates/mangos/13983_01_mangos_spell_template.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE db_version CHANGE COLUMN required_13982_01_mangos_event_ai required_13983_01_mangos_spell_template bit;
+
+UPDATE `wotlkmangos`.`spell_template` SET `EffectSpellClassMask1_1` = 4097 WHERE id IN (53215, 53216, 53217);
```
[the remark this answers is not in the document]
killerwife: @cyberium changed it in the actual commit.

src/game/Spells/Spell.cpp
killerwife: No. We decided against adding it, because it uses reinterpret_cast and we deemed it unnecessary.

PR #270, opened 2019-03-04
MYSQL80 TSQL updates

Cataldo71:
I was getting errors during startup due to use of tsql keywords in the SQL queries. the table name 'groups' and the field 'rank' were both flagging as invalid sql and throwing errors. I simply scoped those to the db/table so that they would resolve properly.

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/sunken_temple/instance_sunken_temple.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/sunken_temple/instance_sunken_temple.cpp
@@ -165,6 +165,10 @@ void instance_sunken_temple::SetData(uint32 uiType, uint32 uiData)
             }
             m_auiEncounter[uiType] = uiData;
             break;
+        case TYPE_MALFURION:
+            if (uiData == IN_PROGRESS)
+                m_auiEncounter[uiType] = uiData;
+            break;
         case TYPE_AVATAR:
             if (uiData == SPECIAL)
             {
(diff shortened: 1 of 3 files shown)


Fabi:
you probably want to use  \`name` instead

Cataldo71:
I tried that with both single/double quotes and tsql didn't seem to like it. technically i should be surrounding the col names with [ ] so maybe ill do that instead, but that doesnt seem to work with the groups table. I cannot find any reference to 'groups' being a protected keyword, but it seems that mysql 8 wants it scoped.

Fabi:
But TSQL is a microsoft thing and not mysql?! I don't think mysql can handle [] at all

Cataldo71:
sorry, i typed tsql meaning mysql.... I tried [ ] and you are correct, it is not recognized by mysql. Scoping the column to the table works just fine as does single quotes.

src/game/Globals/ObjectMgr.cpp
```diff
@@ -3877,7 +3877,7 @@ void ObjectMgr::LoadGroups()
                  "(SELECT COUNT(*) FROM character_instance WHERE guid = group_instance.leaderGuid AND instance = group_instance.instance AND permanent = 1 LIMIT 1), "
                  // 7              8
                  " groups.groupId, instance.encountersMask "
-                 "FROM group_instance LEFT JOIN instance ON instance = id LEFT JOIN groups ON groups.leaderGUID = group_instance.leaderGUID ORDER BY leaderGuid"
+                 "FROM group_instance LEFT JOIN instance ON instance = id LEFT JOIN wotlkcharacters.groups ON groups.leaderGUID = group_instance.leaderGUID ORDER BY leaderGuid"
```
Dramacydal: i guess wotlkcharacters is your local database name, i can be different for the others

cyberium:
Something have to be done for mysql8 but not what is done here

Phatcat:
something was done; https://github.com/cmangos/mangos-wotlk/commit/6a86a0be331f2e42253bdf52a37caf5bd06a4d01 - looks like I missed `` around groups in src/game/Maps/MapPersistentStateMgr.cpp

PR #276, opened 2019-06-08
Pass along player owner to SavePetToDB

Phatcat:
Should hopefully fix the locking error occuring when using the command .saveall

LoadPetFromDB looks to see if the owner is a player towards the end of the function, which in of itself seems ridiculous considering owner in this context is a player type object, not a unit type, and if the owner wasn't an actual player it would have returned false at the beginning where it checks the character database. (there's also some typecasting player into player which is just weird)

Everywhere the pet is saved to the db the owner is checked to be a player first, so the checks inside the function itself are redundant if we can pass along the owner in the function call.

Player.cpp only has 1 instance of SavePetToDB (this is the 1 instance which causes the issue). Easy enough to pass on itself in this case.

Spelleffects has a few instances, but all of them are related to a player caster so we just pass along what we got.

Specifically in EffectSummonPet which I helped to write there is quite a bit of redundant typecasting, it's done everry single time we wanna do something in related to the player owner of the pet, so I've replaced all of that with a single static cast and used that everywhere in that function instead.

Port of cmangos/mangos-classic#390 and https://github.com/cmangos/mangos-tbc/pull/310

### DIFF
--- a/src/game/Entities/Pet.cpp
+++ b/src/game/Entities/Pet.cpp
@@ -366,16 +366,14 @@ bool Pet::LoadPetFromDB(Player* owner, uint32 petentry /*= 0*/, uint32 petnumber
     owner->SetPet(this);                                    // in DB stored only full controlled creature
     DEBUG_LOG("New Pet has guid %u", GetGUIDLow());
 
-    if (owner->GetTypeId() == TYPEID_PLAYER)
-    {
-        ((Player*)owner)->PetSpellInitialize();
-        if (((Player*)owner)->GetGroup())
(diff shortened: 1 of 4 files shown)


Phatcat:
@Cyberium - okay, so I tested this in 3 ways, first I summoned a pet, when summoning a pet it goes through SavePetToDB and no problem passing along the owner, so far so good.

Second I tested specifically unsummoning, since I've now nested the SavePetToDB within the if statement checking if there's an owner, again so far so good.

Third, final and most important test I checked if .saveall would hang the server if a player had a pet. It didn't;

image

PR #282, opened 2019-07-06
Fix extractors build on freebsd

metacreo:
Fix extractors build on freebsd

## 🍰 Pullrequest

this patch allow build extractors on freebsd
### Proof

try  to build
$HOME/mangos
$HOME/build
$HOME: cd build
$HOME/build/: cmake ../mangos -DCMAKE_INSTALL_PREFIX=\../server -DBUILD_EXTRACTORS=ON -DPCH=0 -DDEBUG=0 -DBUILD_PLAYERBOT=ON
### Issues

build failure on freebsd with -DBUILD_EXTRACTORS=ON
added if (UNIX AND NOT LINUX)

### How2Test

- local build success after patch

### Todo / Checklist

- [X] None

### DIFF
--- a/contrib/vmap_extractor/CMakeLists.txt
+++ b/contrib/vmap_extractor/CMakeLists.txt
@@ -22,8 +22,16 @@ ADD_DEFINITIONS("-DIOMAP_DEBUG")
 # build setup currently only supports libmpq 0.4.x
 ADD_DEFINITIONS("-DUSE_LIBMPQ04")
 
-if(UNIX)
+if (UNIX)
   ADD_DEFINITIONS("-ggdb")
+endif()
+
+if (UNIX AND NOT LINUX)
+  add_definitions("-Dfopen64=fopen")
+  add_definitions("-Dfseeko64=fseeko")
+  add_definitions("-Dfseek64=fseek")
+  add_definitions("-Dftell64=ftell")
+  add_definitions("-Dftello64=ftello")
 endif()
 
 if (APPLE)


metacreo:
What about this ?

evil-at-wow:
I'll start with a general remark: you opened this PR three days ago, so be a bit patient please and give people time to react. Some of us have a full time job, are traveling for work regularly, have a family, etc. Of course, if you get no reaction in like two weeks, a reminder is probably acceptable.

Anyway, let's get started with the real comments. 
* I'm not sure, but I think macOS is also considered to be UNIX by CMake, so I think this would add the definitions twice on macOS. Can you confirm this, @cala? If that's the case, we should remove the APPLE section.
* I'm kind of wondering why macOS and FreeBSD would need macros to "replace" calls like fseek64 with fseek, and why that fixes anything. As far as I can see the code uses only one of these, fopen64, in vmapexport.cpp:
```
#ifndef _WIN32
        if (FILE* h = fopen64(path, "rb"))
#else
        if (FILE* h = fopen(path, "rb"))
#endif
```
Given the code uses ```fopen```, ```fclose```, ```fseek``` and so on eveywhere else, this makes no sense to me. The real fix is probably to replace this little block of code with the line in the else block (so just using fopen, not fopen64).
* Either way, I don't think setting up macros to replace (C library) calls is the way to go. For one reason because there are more file functions than listed here. In that regard, the already existing APPLE section was doing it wrong as well IMHO. The problem is basically that the code uses standard C functions like ```fopen```, ```fclose```, ```fseek``` and so on, and we perhaps want to be able to use files larger than 2 GB (certainly in the extractor, but in the vmap extractor that's probably not even needed). Anyway, on Linux with glibc, that's not a given, because offset types like off_t are (signed) 32-bit values for 32-bit builds. Enter the mess of adding and replacing interfaces and macros like ```_LARGEFILE_SOURCE```, ```_LARGEFILE64_SOURCE``` and ```_FILE_OFFSET_BITS```. I always have to check the details in the glibc documentation to be sure, but I think that defining ```_FILE_OFFSET_BITS=64``` does the trick on Linux,  and basically replaces the existing interface with a 64-bit one (so fseek is actually fseek64, etc) for 32-bit builds (and is a no-op for 64-bit builds), offset types are 64-bit, and so on and all is well. I don't know about FreeBSD, but I would expect this to work for 64-bit builds, like on Linux, and perhaps only give problems for 32-bit builds. And it looks like setting up ```_FILE_OFFSET_BITS``` can also do the trick there.

So things I'd like you to do:
* Mention some details for your system, in particular whether your FreeBSD version is 64-bit (or not), and if you're building the extractor tools as 64-bit applications as well. Also the exact compile error you got is very useful.
* Check if replacing the code in vmapexport.cp like I mentioned above works for you. If not, what error do you get when building?
* If the above works, check if setting ```add_definitions("-D_FILE_OFFSET_BITS=64")``` rather than your block of macros also works for you, and if not, what error you get.

metacreo:
OK. Thanks.
Today I will try build with different options. and describe given errors

metacreo:
So... I found the way...
Problem can be resolved with OS definition.
I close this #282.
And open new with small change in vmapexport.cpp:295
--#ifndef _WIN32
++#if !defined(_WIN32) && !defined(__FreeBSD__)
Thanks

metacreo:
Finally construction:
`
#ifndef _WIN32
#ifdef __FreeBSD__
        if (FILE* h = fopen(path, "rb"))
#else
        if (FILE* h = fopen64(path, "rb"))
#endif
#else
        if (FILE* h = fopen(path, "rb"))
#endif

`
Should be ok ?

evil-at-wow:
Not that last block, because that's basically the same as what we have now (uses ```fopen``` on Windows, and ```fopen64``` on all other systems, including FreeBSD, which is not working for you).

Your patch in cmangos#286 should work, but I think the code should use ```fopen``` on all platforms, so we can improve this a bit more. I'll test some other builds this weekend and push patches afterwards.

Thanks for testing!

evil-at-wow:
By the way, you don't need to close a PR and open a new one when you want to update a patch. You can just update your local branch, and then force push to update the existing PR.

cala:
> I'm not sure, but I think macOS is also considered to be UNIX by CMake, so I think this would add the definitions twice on macOS. Can you confirm this, @cala? If that's the case, we should remove the APPLE section.

Currently, I'm able to build extractors on macOS straight from the code using CMake, if that helps.
I'm not 100% sure, but I think macOS is considered an UNIX most of the time by CMake unless told otherwise by using APPLE to do macOS specific stuff.

PR #283, opened 2019-07-07
Mysql 8.0 syntax support

metacreo:
## 🍰 Pullrequest

rank reserved word and must be escaped
https://dev.mysql.com/doc/refman/8.0/en/keywords.html


metacreo:
mistake, sorry

PR #295, opened 2020-02-22
Fix Gift of the Harvester

alanman:
Item casts spell 52481 instead of 52479. Cross referenced with sql and wowhead.

## 🍰 Pullrequest
Quest: The Gift that Keeps on Giving
Item: Gift of the Harvester not creating Scarlet Ghouls because it is looking for the wrong spell in code.

### Proof
Item casts this spellID
https://www.wowhead.com/spell=52479/gift-of-the-harvester

### Issues

- None

### How2Test

- Have a DK with the item Gift of the Harvester, 39253
- Attempt several times to turn a Scarlet Miner into a Scarlet Ghoul in the mines.
- Test this before and after the change.

### Todo / Checklist

- [x] None

### DIFF
--- a/src/game/Spells/SpellEffects.cpp
+++ b/src/game/Spells/SpellEffects.cpp
@@ -10933,7 +10933,7 @@ void Spell::EffectScriptEffect(SpellEffectIndex eff_idx)
                     }
                     return;
                 }
-                case 52479:                                 // Gift of the Harvester
+                case 52481:                                 // Gift of the Harvester
                 {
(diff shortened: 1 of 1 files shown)


killerwife:
This seems to be missing some steps inbetween. Both steps are supposed to be utilized but something is missing somewhere. @Grz3s can u dig into sniff pls?

Grz3s:
quest was made in sd2 ... and was working correct before...
will look whats wrong ...

xfurry:
@Grz3s asked me to close this.
I suspect it's already implemented and working.

PR #299, opened 2020-04-25
Missing spells added

Grz3s:
TDB needs them also


Grz3s:
part of:
https://github.com/cmangos/issues/issues/2178

PR #321, opened 2021-01-18
AHBot fixes and balancing

canidae:
## 🍰 Pullrequest
These changes fixes a crash I can't really explain why happens in the first place (did not occur in Classic core), and it improves item balance as well as adding some items that were missing for WotLK core. It also prevents certain items (notably arrows/bullets) from being valued extremely high, which made engineering unreasonably lucrative .
Changes should be compatible with Classic core and presumably (not tested) TBC core as well.

As for the crash, I captured the following stack trace (not a debug-build):
```
Thread 12 "mangosd" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fed2d481700 (LWP 1165)]
0x0000557a4f2c32f5 in Item::GetProto() const ()
#0  0x0000557a4f2c32f5 in Item::GetProto() const ()
#1  0x0000557a4f1fdb08 in AuctionHouseBot::Update() ()
#2  0x0000557a4f65a1b2 in World::Update(unsigned int) ()
#3  0x0000557a4f1cdd3f in WorldRunnable::run() ()
#4  0x00007fed31287d84 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007fed31d88609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#6  0x00007fed30f75293 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```
There's only one call to `Item::GetProto()` within `AuctionHouseBot::Update()` and there was no check for `nullptr` before invoking the method. I don't know how `Item* item = sAuctionMgr.GetAItem(auction->itemGuidLow);` could result in a `nullptr` when we're iterating `auctionHouse->GetAuctionsBounds()`, but in any case, the extra check seems to prevent crashes.

### Proof
- None

### Issues
- Might fix https://github.com/cmangos/issues/issues/2349

### How2Test
- None

### Todo / Checklist
- [X] None

### DIFF
--- a/src/game/AuctionHouseBot/AuctionHouseBot.cpp
+++ b/src/game/AuctionHouseBot/AuctionHouseBot.cpp
@@ -193,13 +193,13 @@ void AuctionHouseBot::Update()
         }
 
         // remove items we've overridden (AddChance > 0) and add using given AddChance and stack size
-        for (auto itemData : m_itemData)
+        for (auto& itemData : m_itemData)
         {
             if (itemData.second.AddChance > 0) // replace normal loot sources with custom chance of adding item
                 itemMap[itemData.first] = urand(0, 99) < itemData.second.AddChance ? urand(itemData.second.MinAmount, itemData.second.MaxAmount) : 0;
         }
 
-        for (auto itemEntry : itemMap)
+        for (auto& itemEntry : itemMap)
         {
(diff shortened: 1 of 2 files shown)


canidae:
I will update this branch with some further improvements in the future. Merge can wait until then.

cyberium:
Merged thanks

PR #340, opened 2021-05-16
Add playerbot setting for forced running

Nibato:
## 🍰 Pullrequest
Currently the bots walk to loot/collect objects or talk to npcs. I added an config option that forced them to run to it instead.

### Proof
- None

### Issues
- None

### How2Test
- Add a bot
- Have them loot a corpse or object


cala:
Thank you for this report. Bots should run when looting (unless close to the corpse they need to loot) and this was still the case not so long ago (about a few months).

I will test and review your PR soon, but I think the config part won't be needed as walking is not intended in the first place.

cala:
Merged in https://github.com/cmangos/mangos-wotlk/commit/16638fa7c9d1e5281c94c1c65d260edeea44b4d0
Thank you very much. As indicated previously, I dropped the config part because bots are intended to run when looting.

PR #341, opened 2021-05-19
Fixed disenchantment rolls not giving out items

Nibato:
## 🍰 Pullrequest
I created an additional disenchanting case for one of the Loot() constructors. This allows the loot object to have it's owner set when creating the temporary loot object for disenchanting rolls.

### Proof
- None

### Issues
- Fixes #2463 https://github.com/cmangos/issues/issues/2463

### How2Test
- Be in a group with an enchanter
- Roll disenchant on the item and win the roll

### DIFF
--- a/src/game/Loot/LootMgr.cpp
+++ b/src/game/Loot/LootMgr.cpp
@@ -862,11 +862,10 @@ void GroupLootRoll::Finish(RollVoteMap::const_iterator& winnerItr)
             {
                 m_lootItem->isReleased = true;
                 m_lootItem->allowedGuid.clear();
-                ItemPrototype const* pProto = sObjectMgr.GetItemPrototype(m_lootItem->itemId);
-                ItemPosCountVec dest;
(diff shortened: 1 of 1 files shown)


killerwife:
Nice, please check one usecase. Last year i added one experimental packet, if you have the window opened and someone doesnt have space in bags it should go to the mob, and the person having the window opened should get hotswap window packet for this loot. This packet SMSG_LOOT_UPDATE is supposed to do it and thats what i added use of, can u check if that works with your patch?

Nibato:
It seems to work as it should. 

I performed the following steps for your suggested test case:
- Dual boxed two characters
- Filled the inventory with linen cloth
- Opened corpse loot with both characters before rolling
- Rolled disenchant on them both
- The disenchanted materials replaced the original item in the loot window for both players

Proof: https://www.youtube.com/watch?v=PKnYr7VM6Tw

killerwife:
I am happy and merging the PR.

killerwife:
Should you wanna do any more wotlk work, which we direly need, hit me up on discord, I will provide you any resources you might need and I might have.

PR #344, opened 2021-06-12
BroadcastText 30511 corrected

Grz3s:
source: sniff & vid


MantisLord:
We do this as a regular DB update now after https://github.com/cmangos/wotlk-db/commit/3f8cf1438d6ad97e264673906d90aad8edd5801a

PR #352, opened 2022-02-03
Make Auras Build Again

insunaa:
## 🍰 Pullrequest

Fixes Compile for WotLK
std::remove_if uses move semantics, which don't work with const pointers.
This *buillds* but I don't know if it *works* because I can't test WotLK. Please someone test this :)

### Todo / Checklist

- [ ] Build and run server. Try to use runes on a DK and see if it crashes because of some big oversight I made


insunaa:
kek I already see that I forgot to increment the itr. brb

killerwife:
Incorporated as part of https://github.com/cmangos/mangos-wotlk

PR #355, opened 2022-02-13
Show quest greeting only if player has more than one menu option available

Nytanath:
## 🍰 Pullrequest

Currently, quest greetings will be shown even if the player has only one menu option available (by menu option I mean gossip options and also quests available to accept or already accepted).
The easiest NPC to test this with is [Deputy Willem](https://tbc.wowhead.com/npc=823/deputy-willem) in the Human starting zone.
On retail, when you talk with him for the first time, you only have one menu option available: a quest. [He skips directly to that quest](https://youtu.be/UaH5NhxzkSo?t=316), without showing his quest greeting, while in CMaNGOS it shows the quest greeting.
As seen [here](https://youtu.be/UaH5NhxzkSo?t=2009), he only shows his quest greeting when the player can accept two quests from him.
This behavior is the same across every NPC that has a quest greeting.

This is also valid for Classic and TBC.

### Proof

- Videos linked in the description, retail.

### Issues

- None

### How2Test

- Create a Human character, talk with Deputy Willem.


killerwife:
This is actually a modern client feature where it autogoes onto a quest. For gossips with quests it already works like this in tbc client with identical packets as vanilla client, which is sad because you never get to see many of them. So if we want to change this, I need to see a confirmed questgiver_greeting on an npc (only few actually use questgiver_greeting) and video from wrath.

Nytanath:
> which is sad because you never get to see many of them
I agree, but if that's how it was then one has to choose between accuracy and lore :P.
Would it be good to make it a configurable value?

As for proof, I found those old videos: [here the NPC has only one quest](https://youtu.be/GaFoQkueHTE?list=PL33FED3CA4DBF87BD&t=75), and [here there are two](https://youtu.be/SH3HhIsDZ4k?list=PL33FED3CA4DBF87BD&t=442).

killerwife:
Ok I went ahead and scoured the umbral planes for SMSG_QUESTGIVER_QUEST_LIST and I can agree with this. You marked it as draft however so I cannot merge it.

Nytanath:
Added it as configurable value with official behavior as default, I can see why people would prefer it this way even if not strictly retail-like.

killerwife:
I am going to safely assume this is also needed on tbc. But definitely not needed on vanilla, since vanilla even shows single quest gossips.

Nytanath:
> definitely not needed on vanilla, since vanilla even shows single quest gossips

Sorry to reopen this discussion, but I just stumbled on some proof that this behavior is valid for Vanilla too.

[In this video](https://www.youtube.com/watch?v=OAA86_GuQ7E&t=7316s) Sergra Darkthorn is a questgiver with quest greeting, but she does not show it because the player only has one menu item (the quest Echeyakee).

Note that this has nothing to do with the "skip gossip" that was implemented from late WotLK onwards, this is a mechanic that only affects questgivers without gossip flag (which are only a small subset, hence why gathering this data is not easy).

killerwife:
fine with proof as usual

PR #356, opened 2022-03-07
Add config option to force creatures to show their gossip text in all cases

Nytanath:
## 🍰 Pullrequest

Sometime in late TBC (around 2.4 I think?) Blizzard made the client skip a creature's gossip text and pick the gossip option automatically in some cases (mostly class and profession trainers, when they have only one gossip option available to the player).

For example:
.go creature id 5482 (Alliance)
.go creature id 3175 (Horde)

When you talk with one of those NPCs, they skip their gossip text and take you straight to their training windows.

Adding CREATURE_TYPEFLAGS_FORCE_GOSSIP (name makes sense to me, not sure if it has to come from somewhere though) to those creatures makes them show their gossip text and single gossip option instead.

This flag belongs to creature info and is cached by the client, so to see changes you need to drop the cache.

I could not find anything similar to get this behavior on TBC, so this is WotLK-only for the time being.

### Proof

- None

### Issues

- None

### How2Test

- Talk to one of the creatures above, shut down server, activate this option, delete cache, talk with them again.


killerwife:
image

CreatureTypeFlags are a curated part of CreatureStaticFlags because the enum grew too big to send as part of one uint32 in 1.10 i think, so they started only sending those that are valid for client.

Nytanath:
For anyone stumbling on this and interested, I've made an [addon for TBC](https://github.com/Nytanath/ForceGossip-TBC) and [one for WotLK](https://github.com/Nytanath/ForceGossip-WotLK) to get this behavior clientside.

PR #364, opened 2022-06-10
fix death pact targeting returning early error

brz-admin:
fix https://github.com/cmangos/issues/issues/2959

not sure it's the best way to do this tho, it's stopping the spell here because of SPELL_EFFECT_DUMMY having TARGET_UNIT_CASTER as implicitTargetA, so skiping this using spell Id works fine, but might be a bit too brutal ?

### DIFF
--- a/src/game/Spells/Spell.cpp
+++ b/src/game/Spells/Spell.cpp
@@ -5605,7 +5605,8 @@ SpellCastResult Spell::CheckCast(bool strict)
                 }
             }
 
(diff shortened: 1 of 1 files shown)


killerwife:
When you are casting Death Pact, are you targeting someone, self, ghoul or nothing?

brz-admin:
Before this fix : any target says Invalid target.
After : you need to target a ghoul

brz-admin:
Code below that seems to imply you SHOULD target a ghoul but I just made it so it goes through spell specific check

Can't link the lines right now

PR #371, opened 2022-07-22
Shaman: Respect Totem cooldowns even when using a Call spell

insunaa:
## 🍰 Pullrequest

This PR ensures that the "Call of the ..." Spells cannot be used to circumvent Totem cooldowns.

- fixes https://github.com/cmangos/issues/issues/2800


killerwife:
FYI totems in wotlk have smth called SPELL_ATTR_EX7_CAN_BE_MULTICAST which I am guessing is for doing what our TRIGGERED_IGNORE_CURRENT_CASTED_SPELL does. I will retest it myself just for sanity however I find nothing wrong with this change.

killerwife:
Its fine for now, but will need to verify on wotlk whether it should throw oom message or spell not ready yet, when some of the totems are not yet ready.

PR #379, opened 2022-08-27
BossAI: Convert Anub'Rekhan to BossAI+Spell Lists

insunaa:
## 🍰 Pullrequest

This PR converts Anub'Rekhan to BossAI and Spell Lists

### Todo / Checklist

- [ ] Wait for corresponding WotLK-DB PR

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/northrend/naxxramas/boss_anubrekhan.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/northrend/naxxramas/boss_anubrekhan.cpp
@@ -21,7 +21,7 @@ SDComment: Intro text usage is not very clear. Requires additional research.
 SDCategory: Naxxramas
 EndScriptData */
 
-#include "AI/ScriptDevAI/base/CombatAI.h"
+#include "AI/ScriptDevAI/base/BossAI.h"
 #include "AI/ScriptDevAI/include/sc_common.h"
 #include "naxxramas.h"
 enum
--- a/src/game/AI/ScriptDevAI/scripts/northrend/naxxramas/boss_anubrekhan.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/northrend/naxxramas/boss_anubrekhan.cpp
@@ -21,24 +21,24 @@ SDComment: Intro text usage is not very clear. Requires additional research.
 SDCategory: Naxxramas
 EndScriptData */
 
+#include "AI/ScriptDevAI/base/CombatAI.h"
 #include "AI/ScriptDevAI/include/sc_common.h"
 #include "naxxramas.h"
-
 enum
 {
-    SAY_GREET                   = -1533000,
-    SAY_AGGRO1                  = -1533001,
-    SAY_AGGRO2                  = -1533002,
-    SAY_AGGRO3                  = -1533003,
-    SAY_TAUNT1                  = -1533004,
-    SAY_TAUNT2                  = -1533005,
-    SAY_TAUNT3                  = -1533006,
-    SAY_TAUNT4                  = -1533007,
-    SAY_SLAY                    = -1533008,
+    SAY_GREET                   = 13004,
+    SAY_AGGRO1                  = 13000,
+    SAY_AGGRO2                  = 13002,
+    SAY_AGGRO3                  = 13003,
+    SAY_TAUNT1                  = 13006,
+    SAY_TAUNT2                  = 13007,
+    SAY_TAUNT3                  = 13008,
+    SAY_TAUNT4                  = 13009,
+    SAY_SLAY                    = 13005,
 
-    EMOTE_CRYPT_GUARD           = -1533153,
-    EMOTE_INSECT_SWARM          = -1533154,
-    EMOTE_CORPSE_SCARABS        = -1533155,
+    EMOTE_CRYPT_GUARD           = 29887,
+    EMOTE_INSECT_SWARM          = 13443,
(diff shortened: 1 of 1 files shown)


src/game/AI/ScriptDevAI/scripts/northrend/naxxramas/boss_anubrekhan.cpp
```diff
@@ -21,127 +21,105 @@ SDComment: Intro text usage is not very clear. Requires additional research.
...
-        m_introDialogue(aIntroDialogue)
+    boss_anubrekhanAI(Creature* creature) : BossAI(creature, ANUBREKHAN_ACTIONS_MAX),
+        m_instance(static_cast<instance_naxxramas*>(creature->GetInstanceData())),
+        m_isRegularMode(creature->GetMap()->IsRegularDifficulty())
     {
-        m_pInstance = (instance_naxxramas*)pCreature->GetInstanceData();
-        m_bIsRegularMode = pCreature->GetMap()->IsRegularDifficulty();
-        m_introDialogue.InitializeDialogueHelper(m_pInstance);
-        m_bHasTaunted = false;
+        m_hasTaunted = false;
```
killerwife, line 72: ykno, u can also set m_hasTaunted the same way

src/game/AI/ScriptDevAI/scripts/northrend/naxxramas/boss_anubrekhan.cpp
```diff
@@ -21,127 +21,105 @@ SDComment: Intro text usage is not very clear. Requires additional research.
...
+        SetDataType(TYPE_ANUB_REKHAN);
+        AddOnAggroText(SAY_AGGRO1, SAY_AGGRO2, SAY_AGGRO3);
+        AddOnKillText(SAY_SLAY);
+        AddCustomAction(ANUBREKHAN_SUMMON, true, [&](){
+            DoCastSpellIfCan(nullptr, SPELL_SUMMON_GUARD);
+        });
         Reset();
```
killerwife, line 79: With the new paradigm, i strongly suggest not calling Reset in constructor anymore. It gets automatically called on spawn in correct order. Do it in the constructor directly if you need it there. This is because in constructor, some virtual functions arent engaged yet. Which is specifically a problem for SetSpellList which calls an AI func. Also, setspelllist should be in db, normal and 25 have own creature_template entry which can each have different spell list. This way the spell set isnt hardcoded in enum.

src/game/AI/ScriptDevAI/scripts/northrend/naxxramas/boss_anubrekhan.cpp
```diff
@@ -21,127 +21,105 @@ SDComment: Intro text usage is not very clear. Requires additional research.
...
-            m_pInstance->SetData(TYPE_ANUB_REKHAN, DONE);
-    }
-
-    void JustReachedHome() override
-    {
-        if (m_pInstance)
-            m_pInstance->SetData(TYPE_ANUB_REKHAN, FAIL);
-
-        DoCastSpellIfCan(m_creature, SPELL_DOUBLE_ATTACK, CAST_TRIGGERED | CAST_AURA_NOT_PRESENT);
+        BossAI::KilledUnit(pVictim);
```
killerwife, line 107: I strongly suggest calling parent func before your stuff. This is more of a finnicky picky thing, but if blizzard uses generic scripting mechanisms, then events are prolly executed just like in EAI fashion after default behaviour.

src/game/AI/ScriptDevAI/scripts/northrend/naxxramas/boss_anubrekhan.cpp
```diff
@@ -21,127 +21,105 @@ SDComment: Intro text usage is not very clear. Requires additional research.
...
+    }
+
+    void Aggro(Unit *who) override
+    {
+        if (m_isRegularMode)
+            ResetTimer(ANUBREKHAN_SUMMON, 20s);
+        BossAI::Aggro(who);
     }
 
     void KilledUnit(Unit* pVictim) override
```
killerwife, line 99: If you can be bothered, get rid of the hungarian notation like pVictim in any script which you rewrite, so we have a clear progression of consistency. You are already nuking the whole file basically in terms of history, so this is a perfect opportunity to use normal camelCase notation instead. Same goes for instance script.

src/game/AI/ScriptDevAI/scripts/northrend/naxxramas/boss_anubrekhan.cpp
insunaa: I know, but I like setting variables that are used in multiple scripts in the format above, and unique variables in the format below

src/game/AI/ScriptDevAI/scripts/northrend/naxxramas/boss_anubrekhan.cpp
insunaa: If it's called automatically then it can be deleted, but sometimes it seems stuff like the Double Attack spell aren't cast

src/game/AI/ScriptDevAI/scripts/northrend/naxxramas/boss_anubrekhan.cpp
insunaa: Understood

killerwife:
Added the missing spell, please adjust the comment and make a SQL PR. Will make this the first boss to be merged (make sure it adds the spell_target_position for the serverside I added)

PR #395, opened 2022-08-29
BossAI: Convert Gothik to BossAI

insunaa:
## 🍰 Pullrequest

This PR converts Gothik to BossAI and enables future integration with Spell Lists
It would ideally be overhauled to make use of vanilla spells, but their timing and locations are not correct

### Todo / Checklist

- [ ] Wait for corresponding WotLK-DB PR


insunaa:
Something broke somewhere here. Do not merge

PR #399, opened 2022-09-16
Zul'Drak - Spells Related to q.12512 'Leave No One Behind'

Grz3s:
- summoned will follow Player
- Bleeding Out spells will ignore Evade
- All SPELL_ATTR_SS_IGNORE_EVADE spells moved to one section


sql/base/dbc/cmangos_fixes/Spell.sql
```diff
@@ -3070,6 +3044,28 @@ INSERT INTO spell_template(Id, Category, Dispel, Mechanic, Attributes, Attribute
 -- should summon only one creature instead of 4 to 7
 UPDATE spell_template SET EffectDieSides1 = 0, EffectBasePoints1 = 0 WHERE id = 51597;
 
+
+-- ============================================================
+-- SPELL_ATTR_SS_IGNORE_EVADE - IsSpellRemovedOnEvade()
+-- ============================================================
+UPDATE `spell_template` SET `AttributesServerSide` = `AttributesServerSide`|0x00000004 WHERE `Id` IN (
```
killerwife, line 3051: I get that you think this approach is good but this will cause massive merge errors for me. There has be its own section in each section. But sure, they can be centralized.

PR #419, opened 2022-12-11
Shaman: Don't add heroism/bloodlust on sated targets

insunaa:
## 🍰 Pullrequest

This PR makes it so Heroism/Bloodlust isn't cast on characters with the Sated/Exhausted debuff.
There is a minor issue with how it currently works, and that is that when everybody in the group/raid has the sated/exhausted debuff and the shaman uses bloodlust/heroism no sound is being played. I'm not sure if that's how it's supposed to be.

### Issues

- fixes https://github.com/cmangos/issues/issues/3083

### How2Test

- Create Shaman
- Use `.level 79`
- Use `.learn all_myclass`
- Cast Bloodlust/Heroism from the Enhancement Spellbook
- Use `.cooldown clear`
- Cast Bloodlust/Heroism again

### Todo / Checklist

- [x] Research the sound issue


insunaa:
No sound is canon.

PR #425, opened 2022-12-21
Warlock: Fix Seed of Corruption not exploding when stunned or dead

insunaa:
## 🍰 Pullrequest

This PR allows Seed of Corruption to explode even when the Warlock is stunned or dead or LOS, etc.
Additionally this PR allows the explosion from Seed of Corruption to hit the original target in WotLK, as that is intended behaviour.

### How2Test

- Create Warlock
- Learn Seed of Corruption
- Cast Seed of Corruption on enemy
- Cast any Stun spell on self
- Watch what happens with the seed


killerwife:
I will merge this myself, however in the future, stuff like adding TRIGGERED_IGNORE_CASTER_AURA_STATE has to be its own commit. I am lazy with it at times myself, but the issue is that TRIGGERED_IGNORE_CASTER_AURA_STATE belongs in classic core too but seed of corruption code doesnt :)

killerwife:
https://github.com/cmangos/mangos-wotlk/commit/d855a619c6ba06f148196949844f3d2ba6314c07

PR #434, opened 2023-01-27
Fix Wheaty not outputting dumps

Niam5:
## 🍰 Pullrequest
Wheaty Exception Report was not outputting files on crash after repository reorganization.

### Proof
Crash server before applying, no file output

### Issues
- None

### How2Test
Crash server, you get a report now.

### Todo / Checklist
- [X] None


Shauren:
I confirm this indeed fixes the issue as its exactly the same workaround we are using in TrinityCore

killerwife:
Silly, but I will take it.

PR #435, opened 2023-01-28
Naxx: Add missing Spell Scripts and fix minor issues

insunaa:
## 🍰 Pullrequest

This PR adds missing SQL for spell scripts and adds minor fixes such as doors not opening/closing and respawn delays using wrong time measurements


xfurry:
hey
why change the way the doors are working?
```DoUseDoorOrButton``` already takes care of switching the state of the doors based on the instance data sent to the function?

killerwife:
Because its ambiguous. There are doors that lockpick and this paradigm basically bricks them. We should always know what state we are setting them to.

xfurry:
I don't think you can lockpick a door in the middle of the encounter, but I might be wrong. But I agree that some doors can be lockpicked in 5 man instances (first boss in Shattered Halls). For these cases it makes sense to handle the doors more precisely.

The logic isn't ambiguous if you know the flow of instance data 🙂 
Encounter data can move from: Not Started -> In Progress -> Fail -> In Progress -> Done
Door is open in DB and changes: Opened -> Closed -> Opened  -> Closed -> Opened

killerwife:
The proposal wasnt accepted because schmoo didnt agree with what I said. Since the shattered halls exist as a case, its better to have everything consistently rather than like this, especially since the meaning of DoUseDoorOrButton also changes based on gameobject.state column, whereas DoUseOpenableObject is specifically coded to account for that. Im not saying we need to retroactively rewrite everything, however this is the less error prone approach. All toggles that arent for player control generally lead to a bugged encounter somewhere.

PR #436, opened 2023-01-31
Vehicles: Use correct coordinate data for seat locations

insunaa:
## 🍰 Pullrequest

This PR adds the ability to parse 3D location information for vehicle attachments/seats.

### Issues

- implements https://github.com/cmangos/issues/issues/3289

### Todo / Checklist

- [x] Investigate the missing link between AttachmentID in the VehicleSeat.dbc and AttachmentID in the .m2 files.
- [ ] Investigate the possibility of selectively extracting 3D data to save space


insunaa:
I'm gonna remove the meson bits once the todo list is resolved and the rest of the PR is ready to merge

contrib/extractor_scripts/ExtractResources.sh
```diff
@@ -73,6 +74,7 @@ then
   USE_AD="1"
   USE_VMAPS="1"
   USE_MMAPS="1"
+  USE_MODELS="1"
```
killerwife, line 77: is this a significant extraction time? cos if not id just do it always like cameras. Also i assume ur extracting all of them and not just the ones needed for vehicles?

src/game/Entities/Vehicle.cpp
```diff
@@ -40,11 +40,42 @@
...
 #include "Server/SQLStorages.h"
 #include "Movement/MoveSplineInit.h"
 #include "Maps/MapManager.h"
 #include "Entities/Transports.h"
 
+constexpr uint8 attachmentLookup(const VehicleSeatEntry* seatEntry)
+{
+    switch (seatEntry->m_attachmentID)
+    {
+        case 0: return 20;
```
killerwife, line 53: This is actually from client?

contrib/extractor_scripts/ExtractResources.sh
insunaa: it's not, takes a minute or so on my system
I am extracting all creature m2

I added it to be default when using extract all, I think that should be enough

src/game/Entities/Vehicle.cpp
insunaa: yes

PR #437, opened 2023-02-05
BossAI: Add automatic closing and opening of entrance and exit doors with delay

insunaa:
## 🍰 Pullrequest

This PR moves the opening and closing of encounter doors out of the instance script into the BossAI script and adds timers. Maybe the timers should also be custom but I don't know how the system actually works so for now it's static.


killerwife:
Check zul aman, most of the bosses there do it. I believe each boss has a different value.

PR #446, opened 2023-04-06
Spell: Implement Script for Spell 45742

Grz3s:
Req for q.11670 'It Was The Orcs, Honest!'

45742 - Plant Warsong Banner
- only one target
- player must have aura 45760

Target cast 45744 on spell hit

Ref:
https://github.com/cmangos/issues/issues/3343


killerwife:
I might be wrong here, but doesnt this make it castable on a living magmothregar?

PR #450, opened 2023-06-17
Ulduar: Fix Damage calculation delay on projectiles

insunaa:
## 🍰 Pullrequest

This PR fixes an artificial delay in the damage calculation of a projectile.

### How2Test

- Get into a vehicle that can shoot projectiles
- Shoot a projectile at a distant target
- Observe the time it takes for the damage to be applied before and after the patch

### DIFF
--- a/src/game/Spells/Spell.cpp
+++ b/src/game/Spells/Spell.cpp
@@ -1293,12 +1293,13 @@ uint64 Spell::CalculateDelayMomentForDst() const
         }
     }
 
-    return 0;     
+    return 0;
 }
 
 void Spell::RecalculateDelayMomentForDest()
 {
     m_delayMoment = CalculateDelayMomentForDst();
+    m_destTargetInfo.timeDelay = m_delayMoment;
     m_trueCaster->m_events.ModifyEventTime(m_spellEvent, GetDelayStart() + m_delayMoment);
 }
 


killerwife:
This cant be the fix, because CMSG_ON_MISSILE_TRAJECTORY_COLLISION sends completely new XYZ and the delay may radically change due to the different position. Check the TC code on this, I ported it from there.

insunaa:
Test it in game, this is how it works correctly...
With how it was before the delay was doubled each time, because the CMSG_ON_MISS... is sent on impact, but we then add the time to this point of impact to the original amount of time that it already took to impact

src/game/Spells/Spell.cpp
```diff
@@ -1299,6 +1299,7 @@ uint64 Spell::CalculateDelayMomentForDst() const
 void Spell::RecalculateDelayMomentForDest()
 {
     m_delayMoment = CalculateDelayMomentForDst();
+    m_destTargetInfo.timeDelay = 0;
```
killerwife, line 1302: Based on my code analysis, m_destTargetInfo.timeDelay = m_delayMoment;

It designates the time dest effects should go off, GetDelayStart does not change.

insunaa: timeDelay is an independent timer that ticks down once the Event that's set a line below fires. it goes through a while timeDelay >= diff loop
so when that delay is greater than 0 you will not only have the delay that comes from the event, but then also the additional delay that comes from this variable on top

PR #451, opened 2023-06-17
Vehicles: Ensure vehicle attachments stay attached

insunaa:
## 🍰 Pullrequest

This PR ensures that vehicle attachments that have a higher visibility range than the vehicle they're attached to don't appear before the vehicle and thus fall to the ground and appear to not be attached clientside.


killerwife:
I am not sure a case like this should ever happen. Also, this is not the right fix sadly. This function is one of the most heavily executed pieces of code there is, just like all visibility code. I do not think attachments should have a diff vis dist than the vehicle and they should appear at the same time as the vehicle.

insunaa:
I was heavily inspired by TC with this code. Unfortunately the case that attachments spawn before the vehicle itself happens frequently, so at least *something* has to be done about it

PR #454, opened 2023-06-17
Movement: Improve handling of hover movement type

insunaa:
## 🍰 Pullrequest

This PR improves movement when hovering by ensuring future waypoints are offset by the hover height

It also fixes incorrect walkmode/flying flags and fixes spell targeting for hovering creatures.

### DIFF
--- a/src/game/MotionGenerators/TargetedMovementGenerator.cpp
+++ b/src/game/MotionGenerators/TargetedMovementGenerator.cpp
@@ -494,7 +494,6 @@ bool ChaseMovementGenerator::DispatchSplineToPosition(Unit& owner, float x, floa
         this->i_path = new PathFinder(&owner);
 
     bool gen = false;
-    z += owner.GetHoverOffset();
     if (owner.IsWithinDist3d(x, y, z, 200.f) && std::abs(owner.GetPositionZ() - z) < 5.f && owner.IsWithinLOS(x, y, z + i_target->GetCollisionHeight()) && !owner.IsInWater() && !i_target->IsInWater())
     {
         this->i_path->calculate(x, y, z, false, true);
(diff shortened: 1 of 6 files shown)


src/game/MotionGenerators/TargetedMovementGenerator.cpp
```diff
@@ -494,6 +494,7 @@ bool ChaseMovementGenerator::DispatchSplineToPosition(Unit& owner, float x, floa
         this->i_path = new PathFinder(&owner);
 
     bool gen = false;
+    z += owner.GetHoverOffset();
```
killerwife, line 497: Ok so, this is kinda gonna fuck you. The way I understand hover, mob normally moves on ground, but is actually displayed X over ground. So pathfinding should pathfind on the floor and then just offset the points. That is doable in UpdateAllowedPositionZ for example and should solve your problems. Check how tc does it, logically our function works the same just with a couple diff calls. Also, should prolly remove hover flag from IsFlying in wotlk only. In tbc hover is actually used for flight.

insunaa: in sniffs at least the waypoint positions are actually above ground

killerwife: Yes, absolutely, but the pathfinding occurs on the ground. Because hover is not flying, hover is copying floor at Z offset.

insunaa: True. I don't know enough about the movement system to add the offset to the proper place, then

killerwife: image

PR #465, opened 2023-09-27
SD2: fiX for summoning spells in Sholazar Basin Pt.2

Grz3s:
51190 - Summon Lafoo
51191 - Summon Jaloot
51192 - Summon Moodle

+ not used sd2 texts removed (left from removed script)


Grz3s:
need more stuff added... will create new one when is ready

PR #466, opened 2023-09-27
Fix channeled spells using the incorrect target for triggered spells

insunaa:
## 🍰 Pullrequest

This PR changes triggered spells to prefer using the target of the original aura over whatever the unit currently has selected / in focus. For example when starting to channel arcane missiles as a player, selecting another unit mid-channel will continue casting at the original target, instead of at the newly selected target.

### Issues

- fixes https://github.com/cmangos/issues/issues/3461

### How2Test

- Create Char
- Level char to 54
- .tele andorhal
- Walk directly backwards and let the Scarlet Invoker attack you

Edit: ok that test example is a bad example as that particular example has been fixed independently already, tho I'm not sure where.

Edit2: ok so it's fixed on WotLK, but not on Vanilla, but vanilla also has the caster as the target for arcane missiles and then deselects all targets entirely... There's more going on then.


insunaa:
needs more research

PR #469, opened 2023-10-01
Northrend/Borean: Add weather setting to q.11961

Grz3s:
- also unused text removed


Grz3s:
REf:
https://github.com/cmangos/wotlk-db/commit/d3ada0146c80e2222b573b20214b2fb5a2b3edcc

PR #470, opened 2023-10-11
Player: Fix Power Sync between Client & Server

insunaa:
## 🍰 Pullrequest

This PR fixes the player power not being synced between client and server. As soon as the player sees the required amount of power to cast a spell on the client, the player should also be able to cast this spell. This PR enables that functionality.

### Proof

- None

### Issues

- fixes https://github.com/cmangos/issues/issues/3084

### How2Test

- Create Rogue
- Go to any attackable mob
- Spam Sinister Strike while keeping an eye on your energy bar

### Todo / Checklist

- [ ] Confirm this is the correct:tm: way to do things
- [ ] Measure performance impact for multiple players (for 1 player the short loop requires between 0µs and 1µs, the long loop between 8µs and 11µs)

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -1610,7 +1610,7 @@ void Player::Update(const uint32 diff)
     {
         m_regenTimer += diff;
         m_powerUpdateTimer += diff;
-        if (m_powerUpdateTimer >= REGEN_TIME_PRECISE)
+        if (m_regenTimer >= REGEN_TIME_PRECISE)
         {
             RegenerateAll(m_regenTimer);
             if (m_powerUpdateTimer >= REGEN_TIME_FULL)
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -1609,7 +1609,8 @@ void Player::Update(const uint32 diff)
     if (IsAlive())
     {
         m_regenTimer += diff;
-        m_healthRegenTimer += diff;
+        if (!IsInCombat())
+            m_healthRegenTimer += diff;
         if (m_regenTimer >= REGEN_TIME_PRECISE)
             RegenerateAll(m_regenTimer);
     }
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -1619,7 +1619,7 @@ void Player::Update(const uint32 diff)
                 data << GetPackGUID();
                 data << uint8(GetPowerType());
                 data << uint32(GetPower(GetPowerType()));
-                SendMessageToSet(data, GetTypeId() == TYPEID_PLAYER);
+                SendDirectMessage(data);
                 m_powerUpdateTimer -= REGEN_TIME_FULL;
             }
         }
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -1610,15 +1610,18 @@ void Player::Update(const uint32 diff)
     {
         m_regenTimer += diff;
         m_powerUpdateTimer += diff;
-        RegenerateAll(m_regenTimer);
-        if (m_powerUpdateTimer >= REGEN_TIME_FULL)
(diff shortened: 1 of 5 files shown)


src/game/Entities/Player.cpp
```diff
@@ -1609,8 +1609,17 @@ void Player::Update(const uint32 diff)
     if (IsAlive())
     {
         m_regenTimer += diff;
-        if (m_regenTimer >= REGEN_TIME_FULL)
-            RegenerateAll(m_regenTimer / 100 * 100);
+        m_powerUpdateTimer += diff;
+        RegenerateAll(m_regenTimer);
```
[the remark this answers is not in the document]
killerwife: You cannot update it every time due to performance reasons. This is completely invalid fix of the situation. Besides, tick can be from 50 to 500, you cant count on any sort of stable tickrate.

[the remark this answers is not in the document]
insunaa: tick doesn't have to be stable, the diff is used for calculation after all.
we have to do the calculations more often than ever 2s tho, as confirmed by sniffs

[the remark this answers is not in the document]
killerwife: Has to be reworked fully using floats including storage and only sent as uint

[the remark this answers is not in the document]
insunaa: Added. (Meson commit is just so I can build the PR for testing. If you feel like accepting the PR I'll rebase it out and squash the other commits)

src/game/Entities/Unit.cpp
```diff
@@ -10414,16 +10419,17 @@ void Unit::SetHealthPercent(float percent)
     SetHealth(newHealth);
 }
 
-void Unit::SetPower(Powers power, uint32 val, bool withPowerUpdate /*= true*/)
+void Unit::SetPower(Powers power, float val, bool withPowerUpdate /*= true*/)
 {
-    if (GetPower(power) == val)
+    if (GetRealPower(power) == val)
```
insunaa, line 10424: with float imprecision it might be best to keep that as
`if (GetPower(power) == uint32(val))`
so that tiny rounding errors don't cause the whole function to run every time

Not sure

src/game/Entities/Player.cpp
```diff
@@ -1609,8 +1609,20 @@ void Player::Update(const uint32 diff)
     if (IsAlive())
     {
         m_regenTimer += diff;
-        if (m_regenTimer >= REGEN_TIME_FULL)
-            RegenerateAll(m_regenTimer / 100 * 100);
+        m_powerUpdateTimer += diff;
+        if (m_regenTimer >= REGEN_TIME_PRECISE)
```
insunaa, line 1613: It's still my opinion that this regen happens more frequently than twice per second, but I don't have better data to confirm, only videos from WotLK classic, which is indeed faster, but also isn't 3.3.5a, so once again: unsure.

insunaa: Technically sending `SMSG_POWER_UPDATE` is superfluous, given that the client will already have the necessary and correct data from just the `SMSG_COMPRESSED_UPDATE_OBJECT`, but in network traffic it has been observed that `SMSG_POWER_UPDATE` is sent every 2s while regenerating power

src/game/Maps/Map.cpp
```diff
@@ -874,7 +874,12 @@ void Map::Update(const uint32& t_diff)
 #endif
 
     // Send world objects and item update field changes
-    SendObjectUpdates();
+    m_clientUpdateTimer += t_diff;
```
insunaa, line 877: this doesn't seem to break anything and *roughly* aligns with packet data I've seen.
I don't know if this is because ABK servers were just way slower with processing ticks or if there is actually some limit like this

killerwife: Object update tick is 400. Didnt manage to slot that in into our system because its potentially dangerous.

insunaa: should I remove it from the PR or should I correct the value?

PR #471, opened 2023-10-14
Item: Implement Val'anyr effect

insunaa:
## 🍰 Pullrequest

This PR implements the effect of Val'anyr:
45s internal cooldown, 10% proc chance, initial aura only procs on real heals and not on overheals, following shield procs proc on all heals and overheals and accumulate until 20k absorb is reached

### Proof

- [Wowpedia](https://wowpedia.fandom.com/wiki/Val%27anyr,_Hammer_of_Ancient_Kings#Proc)

### Issues

- relates https://github.com/cmangos/issues/issues/3580

### How2Test

- Create healing character
- set health and mana really high
- use .damage 100000 on yourself
- heal yourself with val'anyr equipped

### Todo / Checklist

- [x] Check if the absorb actually does correctly stack


src/game/AI/ScriptDevAI/scripts/world/spell_scripts_wotlk.cpp
```diff
@@ -99,11 +99,63 @@ struct BloodReserveEnchant : public AuraScript
...
+        return SPELL_AURA_PROC_OK;
+    }
+};
+
+// 64411 Blessing of Ancient Kings
+struct BlessingOfAncientKings : public AuraScript
+{
+    SpellAuraProcResult OnProc(Aura* aura, ProcExecutionData& procData) const override
+    {
+        if (!IsPositiveSpell(aura->GetId()))
```
insunaa, line 124: snagged that from the original UnitAuraProcHandler.cpp and I'm not sure what it does, so I kept it. purely from the name it doesn't make much sense, tho

al3xc1985:
OMG, This is nice

PR #473, opened 2023-10-18
Add SQLite DB backend support

insunaa:
## 🍰 Pullrequest

This PR adds support for SQLite as a DB backend as an (optional) alternative to MySQL and PostgreSQL.
We already export our Database files in SQLite anyway, so it makes some sense to be able to use them for running the server.

SQLite performs similarly to MySQL, depending on your storage hardware.
SQLite is fully cross-platform.
SQLite is public domain, there are no possible licensing issues.
SQLite does not require any additional software to be installed on the target computer (tho it requires libsqlite3 on the computer that compiles CMaNGOS).
SQLite uses file-system permissions for permission management, there are no internal users that would need to be created, any user with access to the .sqlite files is assumed to have legitimate access, which means that there's no chance of misconfiguring the DB Server environment in some way, because there is no DB Server.
SQLite has cmd clients on all operating systems, and it's also supported by many popular GUI clients (like HeidiSQL).

### Todo / Checklist

- [ ] More extensive testing if interest exists
- [ ] Fixing bugs discovered in testing or static analysis
- [x] Replace uses of steady_clock with system_clock

This PR currently includes my Meson additions, because Meson is just what I use to build CMaNGOS given that it's much more comfortable to use than CMake, but the PR also fully supports CMake. The build option for SQLite is `-DSQLITE=1`

The PR will remain in draft until interest or disinterest has been established, at which point removing my Meson code may be discussed.


src/game/Accounts/AccountMgr.cpp
```diff
@@ -55,9 +55,21 @@ AccountOpResult AccountMgr::CreateAccount(std::string username, std::string pass
...
+    std::tm* now_tm = std::localtime(&now_c);
+
+    char buffer[80];
+    std::strftime(buffer, 80, "%Y-%m-%d %H:%M:%S", now_tm);
+
     bool update_sv = LoginDatabase.PExecute(
-        "INSERT INTO account(username,v,s,joindate) VALUES('%s','%s','%s',NOW())",
-            username.c_str(), v_hex, s_hex);
+        "INSERT INTO account(username,v,s,joindate) VALUES('%s','%s','%s','%s')",
+            username.c_str(), v_hex, s_hex, buffer);
```
evil-at-wow, line 72: An alternative that might be worth looking into is the 'CURRENT_TIMESTAMP' keyword as default value in the table from "stardard SQL". I would assume most databases support this and then we don't have to do the time conversion. Although it's also not a big deal because this is only at account creation anyway.

src/shared/Database/DatabaseSqlite.cpp
```diff
@@ -0,0 +1,392 @@
...
+    if (result != SQLITE_OK) {
+        sLog.outErrorDb("SQL: %s", sql);
+        sLog.outErrorDb("query ERROR: %s", sqlite3_errmsg(mSqlite));
+        return false;
+    }
+
+    return true;
+}
+
+QueryResult* SQLiteConnection::Query(const char* sql)
```
evil-at-wow, line 105: This is one function that changed to return a ```std::unique_ptr``` so adjust this function accordingly (see similar changes in MySQL/PostgreSQL version).

src/game/Accounts/AccountMgr.cpp
insunaa: `CURRENT_TIMESTAMP` also seems to work as a macro:
```
sqlite> SELECT CURRENT_TIMESTAMP;
2023-10-22 06:19:46
```
So I think it could also just be used in an INSERT INTO statement. I'll have to check tho

insunaa:
Used preprocessor statements to replace UNIX_TIMESTAMP() and NOW() wherever required with SQLite-native alternatives (unixepoch() and datetime())
Should now be mergeable without conflicts (aside from the Meson stuff, which also merges without conflict, but may not be desired by some, for some reason)

Karth-Xyver:
Maybe a dumb question, but: could the small syntax differences in queries (like `DELETE FROM` instead of `TRUNCATE TABLE` and `unixepoch()` instead of `UNIX_TIMESTAMP()`) be handled via string replacement [around here](https://github.com/cmangos/mangos-wotlk/pull/473/files#diff-e7aafc927be929edb886c4a72602d3323148c3a9149960f66014247c1d60c15dR98) instead of duplicating queries? Or is there a reason not to?

insunaa:
Queries aren't duplicated, since the changes are caught inside of preprocessor statements which are evaluated at compiletime. String-replacement would be runtime and cost performance

Karth-Xyver:
Oh I didn't mean compiled, I meant on the source code :P it'd be easy to forget to adapt a new/modified query somewhere on the server and have it not work when swapping from one db to the other. String replacement would've been more plug-and-play, but I see the performance hit.

src/game/Accounts/AccountMgr.cpp
```diff
@@ -55,9 +55,15 @@ AccountOpResult AccountMgr::CreateAccount(std::string username, std::string pass
     const char* s_hex = srp.GetSalt().AsHexStr();
     const char* v_hex = srp.GetVerifier().AsHexStr();
 
+#ifdef DO_SQLITE
+    bool update_sv = LoginDatabase.PExecute(
+        "INSERT INTO account(username,v,s,joindate) VALUES('%s','%s','%s',datetime())",
```
deiteris, line 60: Note that `datetime()` will output UTC time, while MySQL's `NOW()` will output date and time depending on the server/connection session timezone.

src/game/Maps/MapPersistentStateMgr.cpp
```diff
@@ -1067,7 +1067,11 @@ void MapPersistentStateManager::InitWorldMaps()
 void MapPersistentStateManager::LoadCreatureRespawnTimes()
 {
     // remove outdated data
+#ifdef DO_SQLITE
+    CharacterDatabase.DirectExecute("DELETE FROM creature_respawn WHERE respawntime <= unixepoch('now')");
```
deiteris, line 1071: `unixepoch('now')` is equivalent to `unixepoch()`, so `'now'` is redundant.

However, I wonder why `UNIX_TIMESTAMP(NOW())` in MySQL was used. If `NOW()` was added to use server's TZ, then its doesn't work this way, it still produces UTC timestamp.
image

insunaa: Yeah, I know they produce the same result, but because I didn't know why UNIX_TIMESTAMP(NOW()) was used here I figured there might be some reason to it and just left it the same

src/game/Accounts/AccountMgr.cpp
insunaa: I don't think that'll be an issue because the SQLite backend doesn't mix-and-match with the MySQL backend, so it's not like we're going to have MySQL style dates in one place but SQLite style dates in another with conflicting info.
The SQLite integration is meant for quick and easy testing (swapping out the entire database by just copying a file) as well as easy portability, like our Release builds, which are currently only released for Windows because you can't just bundle a portable mysql server with a linux build. I mean you can, but try explaining that to someone who downloads the Release build because they don't know how to compile CMaNGOS

deiteris: That's just a note as I'm not sure what is expected. But I think this issue just needs to be addressed in MySQL code by switching to `UTC_TIMESTAMP()` instead of `NOW()` to avoid TZ shenanigans.

insunaa:
Removed all references to Meson, the PR is purely for SQLite now

src/game/Accounts/AccountMgr.cpp
insunaa: You shouldn't migrate from SQLite to MySQL anyway. The SQLite implementation is mostly meant for testing or for being able to distribute pre-built CMaNGOS binaries that work without the hassle of setting up MySQL properly.
Of particular note is that SQLite has much fewer data-types than MySQL, so while you *can* migrate from MySQL to SQLite by squashing data types (for example all integers are 64bit in SQLite) it'll be hard to return those formats to the MySQL types. And if you're going to go through the effort of changing all times in a dumped DB to migrate it to MySQL, you might as well update the dates along with it

cyberium:
Why ubuntu-release.yml? Are you sure about it? 
If you want to keep it you'll have to update it following recent change on our ci files.
You did a great job, this can be merged beside that.

insunaa:
I mostly added it as an example to show that it *could* work, but realistically due to libc shenanigans it probably doesn't make sense to do an ubuntu release build (not unless I do the whole shebang with Flatpack and such, but that would be a major pain for no gain)

I'll remove the release again, I have still have it in a backup branch, just in case

insunaa:
Found a bug :/
Because SQLite has implicit types if we do a `WHERE` for example or a `<=` in a query and the right side is `'100.0000000'` for example, SQLite will interpret that as a string, then remove the decimal dot and then convert it back into a number.
This can be fixed with switching queries that use `'%f'` to `%f`, I'll get to it

src/game/GameEvents/GameEventMgr.cpp
```diff
@@ -763,7 +763,11 @@ uint32 GameEventMgr::Initialize()                           // return the next e
         }
         while (queryResult->NextRow());
 
+#ifndef DO_SQLITE
```
killerwife, line 766: Bug, old code was truncate

killerwife, line 766: Everywhere you are doing ifdef, suddenly you are doing ifndef, change that

PR #477, opened 2023-10-27
Fix crash in Player.cpp due to out-of-bounds array access

Karth-Xyver:
## 🍰 Pullrequest

Just copying the bounds checking from [here](https://github.com/cmangos/mangos-wotlk/blob/master/src/game/Entities/Player.cpp#L22398).

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


insunaa:
I see nothing wrong with this PR. Can be merged, imo.

PR #480, opened 2023-11-14
Clean up feign death state

Foritus:
## 🍰 Pullrequest
I noticed that the Ingvar encounter is broken at the moment: when he "dies" after phase 1, he stays marked as dead and is unlootable even after you kill him again. After a little digging with the debugger it seems like we are not clearing the feign death state correctly. This is my probably-wrong attempt at fixing it (it does fix the issue with Ingvar, but I don't know what unintended consequences it might have 🙂)

One thing to investigate is that the animation when Ingvar gets resurrected seems broken by this feign death mechanic (he just lies on the floor and then suddenly turns into a standing scourged version of himself, the spinny transformation does not happen).

### Proof
image
image

### Issues

Fixes #3537

### How2Test
In Utgarde Keep:
`.npc tempspawn 23954 100000`

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Entities/Unit.h
+++ b/src/game/Entities/Unit.h
@@ -2436,7 +2436,14 @@ class Unit : public WorldObject
         void SetImmobilizedState(bool apply, bool stun = false, bool logout = false);
         ///----------End of crowd control methods----------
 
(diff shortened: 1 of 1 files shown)


killerwife:
Can confirm this is not the correct fix. But good news is i know the fix.

PR #483, opened 2023-11-27
Pet: Keep permanent pet when mounting up

insunaa:
## 🍰 Pullrequest

This PR ensures that permanent mounts can stay summoned if the config option for it is set


src/game/Entities/Player.cpp
```diff
@@ -1925,14 +1925,17 @@ bool Player::Mount(uint32 displayid, const Aura* aura/* = nullptr*/)
         return false;
 
     // Custom mount (non-aura such as taxi or command) or in flight: unsummon any pet
-    if (!aura || IsFreeFlying() || IsSpellHaveAura(aura->GetSpellProto(), SPELL_AURA_MOD_FLIGHT_SPEED_MOUNTED))
-        UnsummonPetTemporaryIfAny();
+    if (!aura && !sWorld.getConfig(CONFIG_BOOL_KEEP_PET_WHEN_MOUNTED))
```
killerwife, line 1928: Pretty sure you broke the bool logic here cos what can happen is you enter this part of the if and you dont do anything. Any flow in this func should either end in "unsummon" or "PET_MODE_DISABLE_ACTIONS"

src/game/MotionGenerators/MovementHandler.cpp
```diff
@@ -349,6 +349,14 @@ void WorldSession::HandleMovementOpcodes(WorldPacket& recv_data)
             mover = _player;
     }
 
+    if (opcode == CMSG_MOVE_SET_FLY)
+    {
+        if (plMover)
+            if (Pet* pet = plMover->GetPet())
+                if (pet->isControlled() && plMover->HasAuraType(SPELL_AURA_MOD_FLIGHT_SPEED_MOUNTED))
```
killerwife, line 356: Cheaper to check ismounted and MOVEFLAG_CAN_FLY

insunaa: I think the moveflag_can_fly is redundant in this case, because `CMSG_MOVE_SET_FLY` already implies that a takeoff is happening, so it can't happen without that moveflag anyway. I'll just check for IsMounted

PR #484, opened 2023-12-20
Expansion-specific XP modifiers

LunaticInAHat:
## 🍰 Pullrequest
This PR allows server admins to set different kill & quest XP modifiers for maps belonging to different expansions. This allows admins to have some broad control over the leveling curve; if they want to level from 0-60 very quickly, and then very slowly 60+, they can now do that.

Perhaps this isn't such a desirable feature to have, since it's not very "Blizz-like", but we already have the ability to set XP modifiers, so I figured that expanding that capability might be interesting. I know I like being able to speed up the endgame a bit, while still being able to spend time in the Vanilla levels on new characters.

### Proof
- None

### Issues
- None

### How2Test
Set different values for the modifiers in the config file. Go to a Vanilla map, kill a monster / complete a quest, and observe that the correct modifier was applied. Rinse & repeat across BC- and Wrath-specific maps.

### Todo / Checklist
- [X] None


al3xc1985:
Nice. I was looking for something like this

kelbren:
I would suggest adding xp  modifier for discovery and pet xp as well.

LunaticInAHat:
> I would suggest adding xp modifier for discovery and pet xp as well.

Done!

src/game/Maps/Map.cpp
```diff
@@ -1260,6 +1260,62 @@ void Map::SetNewDifficultyCooldown(TimePoint const& newCooldown)
         m_dynamicDifficultyCooldown = newCooldown;
 }
 
+float Map::GetPetKillXPMod() const
```
kelbren, line 1263: Expansion() Does this change with what expansion a map is on? I assumed this function returned the value from the account. If that is the case, as far as I can tell, this would only set the xp rate to the expansion setting in the account regardless of the map. I was under the impression that the desired behavior was to have variable rates for different expansion content (maps) ?
Perhaps this change would do what your trying to do?
```
float Map::GetPetKillXPMod() const
{    
    MapEntry const* mapEntry = sMapStore.LookupEntry(GetMapId());
    switch (mapEntry->Expansion())
    {
        default:
        case 0:
            return sWorld.getConfig(CONFIG_FLOAT_RATE_PET_XP_KILL_VANILLA);
        case 1:
            return sWorld.getConfig(CONFIG_FLOAT_RATE_PET_XP_KILL_BC);
        case 2:
            return sWorld.getConfig(CONFIG_FLOAT_RATE_PET_XP_KILL_WOTLK);
    }
} 
```

src/game/Maps/Map.h
```diff
@@ -221,8 +221,13 @@ class Map : public GridRefManager<NGridType>
         void ChangeMapDifficulty(Difficulty difficulty);
         void SetNewDifficultyCooldown(TimePoint const& newCooldown);
         TimePoint const& GetNewDifficultyCooldown() const { return m_dynamicDifficultyCooldown; }
+        float GetPetKillXPMod() const;
+        float GetQuestXPMod() const;
+        float GetKillXPMod() const;
+        float GetExploreXPMod() const;
 
         MapEntry const* GetEntry() const { return i_mapEntry; }
+        uint32 Expansion() const { return (i_mapEntry) ? i_mapEntry->Expansion() : 0u; }
```
insunaa, line 230: @kelbren

src/game/Maps/Map.cpp
LunaticInAHat: `Expansion()` does indeed change with the expansion that a map is from. It does not reference the player's account; it does effectively the same thing as what you suggested. At map load time, `i_mapEntry` will have been fetched from `sMapStore`. Then, calling `Expansion()` on the map will return the expansion value that was stored in `Maps.dbc`. No point in re-querying the DBC, when we've already got a reference to the correct record.

src/game/Maps/Map.h
killerwife: GetExpansion pretty sure. I know the submethod is called Expansion only but we dont have to copy old codestyle :D

src/game/Maps/Map.cpp
killerwife: Player wont be allowed to enter expansion map if he has no rights for given expansion. So correct assumption. Doing a lookup here is meh.

src/game/Maps/Map.h
LunaticInAHat: Makes sense; name changed accordingly

cyberium:
Merged thanks

PR #491, opened 2024-01-14
Spell: Mindflay calculate base damage for tick

insunaa:
## 🍰 Pullrequest

Ensure that the periodic trigger spell has basepoints

### Proof

- None

### Issues

- fixes https://github.com/cmangos/issues/issues/3653

### How2Test

- see issue

### Todo / Checklist

- [x] Might be hackfix? I think periodic triggered spells should generally get the basepoints they're triggered with, this shouldn't need a special case I believe


killerwife:
Check SpellAuras.cpp:2340 why its failing to be filled there.

insunaa:
@killerwife updated

PR #492, opened 2024-01-16
Fix NPC name in new mail notification when receiving mails from NPCs

Karth-Xyver:
## 🍰 Pullrequest

This was fun to track down, even had to parse packets to see where the new mail notification comes from...

If a player receives a mail from a NPC, and that NPC happens to have the same entry as an existing player's guid, then the mail notification would show the player's name instead of the NPC's name.

The packet needs both the highguid and the entry in two separate fields. I tried using either one or the other and it didn't work.

Added it for gameobjects too, since I noticed that type in the mail enum, but did not test it (can gameobjects even send mails?).

### Proof

- None

### Issues

- None

### How2Test

- Create a new character, edit the guid in the database to 16128.
- Level up a character to 80. This causes the NPC Rhonin (NPC entry: 16128) to send the player a mail.
- Hover the mouse on the new mail notification - it won't show Rhonin, it will show the previously created character's name.
This fixes that.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Mails/MailHandler.cpp
+++ b/src/game/Mails/MailHandler.cpp
@@ -836,21 +836,9 @@ void WorldSession::HandleQueryNextMailTime(WorldPacket& /**recv_data*/)
                     break;
             }
 
-            data << ObjectGuid(guidType, m->sender); // sender guid
-
-            switch (m->messageType)
-            {
-                case MAIL_AUCTION:
-                case MAIL_CREATURE:
-                case MAIL_GAMEOBJECT:
-                    data << uint32(m->sender);              // sender's id
-                    data << uint32(m->messageType);         // message type
-                    break;
-                default:
-                    data << uint32(0);
-                    data << uint32(0);
-                    break;
-            }
+            data << ObjectGuid(guidType, m->sender);     // sender guid
+            data << static_cast<uint32>(m->sender);      // sender id
+            data << static_cast<uint32>(m->messageType); // message type
 
             data << uint32(m->stationery);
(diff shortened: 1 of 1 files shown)


src/game/Mails/MailHandler.cpp
```diff
@@ -824,13 +824,27 @@ void WorldSession::HandleQueryNextMailTime(WorldPacket& /**recv_data*/)
...
                 case MAIL_AUCTION:
-                    data << uint32(m->sender);              // auction house id
-                    data << uint32(MAIL_AUCTION);           // message type
+                case MAIL_CREATURE:
+                case MAIL_GAMEOBJECT:
+                    data << uint32(m->sender);              // sender's id
+                    data << uint32(m->messageType);         // message type
                     break;
                 default:
                     data << uint32(0);
```
insunaa, line 850: from my naive reading of the Mail class the `m->sender` and `m->messageType` values should always be defined and shouldn't require an additional switch
I think just adding them like
```cpp
data << static_cast<uint32>(m->sender);
data << static_cast<uint32>(m->messageType);
```
should be fine

Karth-Xyver: I tried to keep the original behavior for the other mail cases, so it would send 0 and 0 when mail is not from auction/character/gameobject. I will test to see if it would break anything to always send data, thanks for the idea!

Karth-Xyver: Tested with all cases, everything works exactly as before so I think it's safe to change.

cyberium:
Merged thank you

PR #493, opened 2024-01-26
This PR fixes a few compiler warnings when compiling with Clang

insunaa:
## 🍰 Pullrequest

Fix or silence a bunch of compiler warnings

### How2Test

- Compile core with Clang and warnings enabled

### Todo / Checklist

- [ ] Check if changes are appropriate

### DIFF
--- a/src/game/BattleGround/BattleGroundEY.cpp
+++ b/src/game/BattleGround/BattleGroundEY.cpp
@@ -162,6 +162,7 @@ void BattleGroundEY::SetTowerOwner(EYNodes node, Team team)
             GetBgMap()->GetVariableManager().SetVariable(WORLD_STATE_EY_DRAENEI_RUINS_NEUTRAL, m_towerOwner[NODE_DRAENEI_RUINS] == TEAM_NONE);
             break;
         }
+        default: break;
     }
 }
 
--- a/src/game/BattleGround/BattleGroundEY.cpp
+++ b/src/game/BattleGround/BattleGroundEY.cpp
@@ -162,7 +162,7 @@ void BattleGroundEY::SetTowerOwner(EYNodes node, Team team)
             GetBgMap()->GetVariableManager().SetVariable(WORLD_STATE_EY_DRAENEI_RUINS_NEUTRAL, m_towerOwner[NODE_DRAENEI_RUINS] == TEAM_NONE);
             break;
         }
-        default: break;
+        case EY_MAX_NODES: break;
     }
 }
 
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -5929,7 +5929,7 @@ void Player::UpdateCombatSkills(Unit* /*pVictim*/, WeaponAttackType attType, boo
         return;
 
     // The farther player is from the cap, the easier it gets to level up the skill
-    float chance = ((float(std::max(1, (room / 5))) / (cap / 5)) * 100);
+    float chance = ((float(std::max(1, (room / 5))) / (cap / 5.f)) * 100);
 
     // Weapon skills: increase chance by intellect
     if (!defence)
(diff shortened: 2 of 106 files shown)


src/game/GameEvents/GameEventMgr.cpp
```diff
@@ -1462,6 +1462,7 @@ void GameEventMgr::ComputeEventStartAndEndTime(GameEventData& data, time_t today
             data.end = mktime(&todayT);
             break;
         }
+        default: break;
```
evil-at-wow, line 1465: Maybe here as well just add the two missing value (cf earlier comment)?

src/game/Globals/CombatCondition.cpp
```diff
@@ -181,6 +181,7 @@ bool CombatConditionMgr::Meets(Unit const* self, int32 Id, float range)
                     case ConditionLogic::AND: if (!results[0] || !results[1]) return false; break;
                     case ConditionLogic::OR:  if (!results[0] && !results[1]) return false; break;
                     case ConditionLogic::XOR: if (results[0] == results[1])  return false; break;
+                    default: break;
```
evil-at-wow, line 184: I'm slightly surprised clang complained here. At least in theory it could have seen that the missing NONE was not possible here considering the surrounding code :smile:

src/shared/Database/DatabaseSqlite.cpp
```diff
@@ -35,13 +35,15 @@ size_t DatabaseSqlite::db_count = 0;
 DatabaseSqlite::DatabaseSqlite()
 {
     // before first connection
-    if (db_count++ == 0);
+    if (db_count++ == 0)
+    ;
```
evil-at-wow, line 39: This was probably inspired by the MySQL class that needed to initialize the MySQL library. As far as I remember SQLite doesn't need anything like that, so I would just drop this from the commit.

And then clean up and remove the whole body (and similar for the destructor) and the ```db_count``` member,  because it's not looking like it serves any purpose anyway...

insunaa: True, I never saw any place aside from the constructor and destructor that access this variable either, I just kept it in case there was something I was missing. I'll remove it

src/game/Entities/Player.cpp
```diff
@@ -5929,7 +5929,7 @@ void Player::UpdateCombatSkills(Unit* /*pVictim*/, WeaponAttackType attType, boo
         return;
 
     // The farther player is from the cap, the easier it gets to level up the skill
-    float chance = ((float(std::max(1, (room / 5))) / (cap / 5)) * 100);
+    float chance = ((float(std::max(1, (room / 5))) / (cap / 5.f)) * 100);
```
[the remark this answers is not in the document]
insunaa: I wasn't sure about this change, so I put it in a separate commit. Clang complains about a possible loss of precision and I agree it's probably not going to occur. I'm going to check if there's another way to silence the integer division compiler warning for this particular calculation

src/game/BattleGround/BattleGroundEY.cpp
```diff
@@ -162,6 +162,7 @@ void BattleGroundEY::SetTowerOwner(EYNodes node, Team team)
             GetBgMap()->GetVariableManager().SetVariable(WORLD_STATE_EY_DRAENEI_RUINS_NEUTRAL, m_towerOwner[NODE_DRAENEI_RUINS] == TEAM_NONE);
             break;
         }
+        default: break;
```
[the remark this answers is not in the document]
insunaa: yup :/

cyberium:
Look ok to me to be merged, but this need to be synced to other core too. Do you want me do to it?

insunaa:
I can do the sync, I just didn't do it ahead of time because I wanted to be sure there's interest in merging this

cyberium:
Alright i dont see anything relevant that have to be addressed so green light from me.

PR #495, opened 2024-01-29
Spell: Fix Dream Vision allowing autohit

insunaa:
## 🍰 Pullrequest

This PR ensures the creature summoned by Dream Vision cannot enter combat in any way

### Issues

- fixes https://github.com/cmangos/issues/issues/3634

### How2Test

- Go into a zone with enemies
- `.additem 9197`
- Use the item
- try to attack enemies while channeling

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/world/item_scripts.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/world/item_scripts.cpp
@@ -400,6 +400,27 @@ struct GoblinBomb : public SpellScript
     }
 };
 
+// 11403 - Dream Vision
+struct DreamVision : public SpellScript
+{
+    void OnSummon(Spell* /*spell*/, Creature* summon) const override
+    {
+        if (summon->GetEntry() != 7863)
+            return;
+
+        summon->SetHover(true);
+        summon->SetWaterWalk(true);
+        summon->SetFeatherFall(true);
+    }
+
+    void OnRadiusCalculate(Spell* /*spell*/, SpellEffectIndex effIdx, bool /*targetB*/, float& radius) const override
+    {
+        if (effIdx != EFFECT_INDEX_0)
+            return;
+        radius = 2.f;
+    }
+};
+
 void AddSC_item_scripts()
 {
     Script* pNewScript = new Script;
@@ -440,4 +461,5 @@ void AddSC_item_scripts()
     RegisterSpellScript<ArgussianCompass>("spell_argussian_compass");
     RegisterSpellScript<SummonGoblinBomb>("spell_summon_goblin_bomb");
     RegisterSpellScript<GoblinBomb>("spell_goblin_bomb");
+    RegisterSpellScript<DreamVision>("spell_dream_vision");
 }
--- a/sql/scriptdev2/spell.sql
+++ b/sql/scriptdev2/spell.sql
@@ -43,6 +43,7 @@ INSERT INTO spell_scripts(Id, ScriptName) VALUES
 (12948,'spell_hakkar_summoned'),
 (13278,'spell_gdr_channel'),
 (13493,'spell_gdr_periodic'),
+(11403,'spell_dream_vision'),
 (11886,'spell_capture_beast'),
 (11887,'spell_capture_hippogryph'),
 (11888,'spell_capture_faerie_dragon'),
--- a/src/game/Entities/CreatureSettings.cpp
+++ b/src/game/Entities/CreatureSettings.cpp
@@ -40,6 +40,19 @@ void CreatureSettings::ResetStaticFlags(CreatureStaticFlags staticFlags, Creatur
         m_owner->GetVisibilityData().SetVisibilityDistanceOverride(VisibilityDistanceType::Gigantic);
     if (HasFlag(CreatureStaticFlags3::INFINITE_AOI))
         m_owner->GetVisibilityData().SetVisibilityDistanceOverride(VisibilityDistanceType::Infinite);
+
+    if (HasFlag(CreatureStaticFlags::UNINTERACTIBLE))
+        m_owner->SetFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_UNINTERACTIBLE);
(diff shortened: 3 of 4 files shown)


insunaa:
Technically the summoned "Dream Vision" NPC should also not be targetable, but if I add the `UNINTERACTIBLE` unit flag it will still be targetable, only instead of of its name it will say "Unknown" so I removed it from the PR again

src/game/AI/ScriptDevAI/scripts/world/item_scripts.cpp
```diff
@@ -364,6 +364,22 @@ struct ArgussianCompass : public AuraScript
     }
 };
 
+// 11403 - Dream Vision
+struct DreamVision : public SpellScript
+{
+    void OnSummon(Spell* spell, Creature* summon) const override
+    {
+        if (summon && summon->GetEntry() == 7863)
```
killerwife, line 372: For future reference, summon always exists

insunaa:
If you think this is alright I'll squash the commits. Leaving the history for now for reference.

killerwife:
Visibility off would mean not showing up in sniffs, thats bad.

Rest is doable in db i think, here is excerpt:

[3] (22) UNIT_FIELD_BYTES_0_1: 0
[3] (22) UNIT_FIELD_BYTES_0_2: 1
[3] (22) UNIT_FIELD_BYTES_0_3: 2
[3] (22) UNIT_FIELD_BYTES_0_4: 1
[3] (23) UNIT_FIELD_HEALTH: 4050
[3] (26) UNIT_FIELD_POWER3: 100
[3] (27) UNIT_FIELD_POWER4: 100
[3] (28) UNIT_FIELD_POWER5: 1000000
[3] (31) UNIT_FIELD_MAXHEALTH: 4050
[3] (34) UNIT_FIELD_MAXPOWER3: 100
[3] (35) UNIT_FIELD_MAXPOWER4: 100
[3] (36) UNIT_FIELD_MAXPOWER5: 1000000
[3] (53) UNIT_FIELD_LEVEL: 70
[3] (54) UNIT_FIELD_FACTIONTEMPLATE: 1
[3] (58) UNIT_FIELD_FLAGS: (50332424) PlayerControlled, NotAttackbleByPlayerControlled, OnlyAttackableByPlayerControlled, PossessedByPlayer, NotSelectable
[3] (59) UNIT_FIELD_FLAGS_2: (2050) Unk1, RegeneratePower
[3] (61) UNIT_FIELD_BASEATTACKTIME_1: 2000
[3] (61) UNIT_FIELD_BASEATTACKTIME_2: 2000
[3] (64) UNIT_FIELD_BOUNDINGRADIUS: 0.5
[3] (65) UNIT_FIELD_COMBATREACH: 1
[3] (66) UNIT_FIELD_DISPLAYID: 6908
[3] (67) UNIT_FIELD_NATIVEDISPLAYID: 6908
[3] (80) UNIT_MOD_CAST_SPEED: 1
[3] (81) UNIT_CREATED_BY_SPELL: 11403
[3] (122) UNIT_FIELD_BYTES_2_1: 1
[3] (122) UNIT_FIELD_BYTES_2_2: 0
[3] (122) UNIT_FIELD_BYTES_2_3: 0
[3] (122) UNIT_FIELD_BYTES_2_4: 0
[3] (146) UNIT_FIELD_HOVERHEIGHT: 1




Movement Flags: Waterwalking, CanSafeFall, Hover (1879048192)

insunaa:
Added the correct flags and removed the invisibility. Currently the summon spawns 100m to the left of the user, which I don't think is right. I'm assuming that's caused by the EffectRadiusIndex1 being 12. afaik the summon is supposed to spawn ~1m next to the user

killerwife:
That radius is for the mob to despawn if player moves more than 100yd from it. Use OnRadiusCalculate hook to supply proper distance to targeting.

killerwife:
Meant to have these static flags in db:

image

So add those to wotlk db, remove the delta from core whats not needed and then we can push it.

src/game/Entities/Unit.cpp
```diff
@@ -12605,11 +12605,11 @@ Unit* Unit::TakePossessOf(SpellEntry const* spellEntry, SummonPropertiesEntry co
 
     const bool immunePC = IsImmuneToPlayer();
     if (possessed->IsImmuneToPlayer() != immunePC)
-        possessed->SetImmuneToPlayer(immunePC);
+        possessed->SetImmuneToPlayer(possessed->GetSettings().HasFlag(CreatureStaticFlags::IMMUNE_TO_PC) ? true : immunePC);
 
     const bool immuneNPC = IsImmuneToNPC();
     if (possessed->IsImmuneToNPC() != immuneNPC)
-        possessed->SetImmuneToNPC(immuneNPC);
+        possessed->SetImmuneToNPC(possessed->GetSettings().HasFlag(CreatureStaticFlags::IMMUNE_TO_NPC) ? true : immuneNPC);
```
insunaa, line 12612: these are sus. should I just keep it in the spellscript as manual override?

killerwife: No, these are entirely separate, this code here shouldnt be altered. The code is literally "preserve what mob had before posses".

insunaa: in practice however this means that the mob from Dream Vision will not have IMMUNE_TO_PC and IMMUNE_TO_NPC unitflags, because with the code here as it was before it will be reset to whatever the player who summons the dream vision NPC has as flags (and the player doesn't get these flags)

insunaa: so either I need these changes here, or I have to separately set these flags in the SpellScript, because those are applied after the possession

killerwife: Removed the offending block entirely after examination. Rebase your thing and make sure the db data makes it there as intended

insunaa:
image
aside from this it looks good now.

killerwife:
Merged by hand.

PR #502, opened 2024-02-16
Config: Implement config override with env vars

walkline:
## 🍰 Pullrequest
Implement overriding of configuration from the .conf file with environment variables. Environment variables keys are autogenerated based on the keys defined in .conf file. 

Usage example:
```
$ export CM_DATA_DIR=/usr
$ CM_WORLD_SERVER_PORT=8080 ./mangosd
```
### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Anticheat/module/libanticheat.cpp
+++ b/src/game/Anticheat/module/libanticheat.cpp
@@ -215,7 +215,7 @@ namespace NamreebAnticheat
 {
 void AnticheatLib::Reload()
 {
-    if (!sAnticheatConfig.SetSource(_LIB_ANTICHEAT_CONFIG, "Mangosd_"))
+    if (!sAnticheatConfig.SetSource(_LIB_ANTICHEAT_CONFIG, "Anticheat_"))
         sLog.outError("[Anticheat] Could not find configuration file %s.", _LIB_ANTICHEAT_CONFIG);
 
     // the configuration setting must be loaded before the database data because the current config settings
--- a/src/game/Anticheat/module/libanticheat.cpp
+++ b/src/game/Anticheat/module/libanticheat.cpp
@@ -215,7 +215,7 @@ namespace NamreebAnticheat
 {
 void AnticheatLib::Reload()
 {
-    if (!sAnticheatConfig.SetSource(_LIB_ANTICHEAT_CONFIG))
+    if (!sAnticheatConfig.SetSource(_LIB_ANTICHEAT_CONFIG, "Mangosd_"))
         sLog.outError("[Anticheat] Could not find configuration file %s.", _LIB_ANTICHEAT_CONFIG);
 
     // the configuration setting must be loaded before the database data because the current config settings
--- a/src/realmd/Main.cpp
+++ b/src/realmd/Main.cpp
@@ -126,7 +126,7 @@ int main(int argc, char* argv[])
         Log::WaitBeforeContinueIfNeed();
         return 1;
     }
-    
+
 #ifndef _WIN32                                               // posix daemon commands need apply after config read
     if (vm.count("s"))
     {
--- a/src/realmd/Main.cpp
+++ b/src/realmd/Main.cpp
@@ -120,15 +120,13 @@ int main(int argc, char* argv[])
     }
 #endif
 
-    if (!sConfig.SetSource(configFile))
+    if (!sConfig.SetSource(configFile, "Realmd_"))
     {
         sLog.outError("Could not find configuration file %s.", configFile.c_str());
         Log::WaitBeforeContinueIfNeed();
         return 1;
     }
     
-    std::vector<std::string> overriddenConfKeys = sConfig.OverrideWithEnvVariablesIfAny();
-
 #ifndef _WIN32                                               // posix daemon commands need apply after config read
     if (vm.count("s"))
     {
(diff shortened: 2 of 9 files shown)


killerwife:
I like the idea, I do not like the implementation. In my opinion, the code should not be conditional and should not do two passes over variables. Loading should be done in a simple form:

1. Load config value in realmd.conf or mangosd.conf
2. Check if corresponding environment variable exists and override it accordingly

The code flow should reflect it and no additional calls to config library such as OverrideWithEnvVariablesIfAny should be necessary. Thats a footgun for a day when someone makes a new binary and forgets to call it.

Also the nomenclature is a footgun, because someone will not read the documentation and not realize it should be underscore delimited, capitalized. We can keep it as is, plenty of programming languages dont have them like that.

However I would suggest to specify a prefix as Mangosd.WorldServerPort to avoid conflict with any same named world vs realm variables (so you can have both running on same metal)

insunaa:
I guess generally when you specify config variables with environment variables like this you don't do that in some hard-coded/system-wide manner, you explicitly set those vars for calling that specific application like
`WORLDPORT=8080 EXPRATE=50 ./mangosd -c [...]`
or in a systemd service file, or in a dockerfile.

Ofc env variables can also be systemwide, but specifically for projects like cmangos I think the point of using them would be to specify them immediately and explicitly for the application call

walkline:
And also, I believe there would be issues with dots (like here Mangosd.WorldServerPort) on some systems, so we would need to escape them anyway.

killerwife:
> Alright, two questions then:
> 
> 1. Should we expose overridden keys so we can show the user in the right place that these keys were overridden with environment variables? Or should we not expose them and instead log them inside the Config at the moment when they were actually overridden? Or should we not log them at all?
> 2. Can you provide more details about the prefix? I didn't get the part about "world vs realm". Do you mean for the mangosd binary we should use the `Mangosd` prefix, and for realmd it should be `Realmd`?

1. No applications i work with mention in logs overriden variables through env vars. So I would not do any logging on the overriden ones myself, but if someone has a valid use to it then I will not really object.
2. yes

walkline:
I made a few changes to the implementation and adjusted the format of environment variable names as suggested (though I prefer screaming snake case more). Now, for configurations like `Rate.Health` and `DataDir`, you need to specify the following environment variables:
```
export Mangosd_Mangosd_Rate_Health=1.2
export Mangosd_DataDir=/tmp
```

src/shared/Config/Config.cpp
```diff
@@ -62,15 +67,20 @@ bool Config::Reload()
...
+        if (envValue)
+            value = *envValue;
+        
         newEntries[entry] = value;
     }
     while (in.good());
 
     m_entries = std::move(newEntries);
-
+
```
killerwife, line 83: Whitespace

src/realmd/Main.cpp
```diff
@@ -120,13 +120,13 @@ int main(int argc, char* argv[])
...
 
-    if (!sConfig.SetSource(configFile))
+    if (!sConfig.SetSource(configFile, "Realmd_"))
     {
         sLog.outError("Could not find configuration file %s.", configFile.c_str());
         Log::WaitBeforeContinueIfNeed();
         return 1;
     }
-
+
```
killerwife, line 129: Whitespace

src/game/Anticheat/module/libanticheat.cpp
```diff
@@ -215,7 +215,7 @@ namespace NamreebAnticheat
 {
 void AnticheatLib::Reload()
 {
-    if (!sAnticheatConfig.SetSource(_LIB_ANTICHEAT_CONFIG))
+    if (!sAnticheatConfig.SetSource(_LIB_ANTICHEAT_CONFIG, "Mangosd_"))
```
killerwife, line 218: Would honestly preface this with Anticheat_

killerwife:
Add a comment on top of each Mangosd.conf.dist, realmd and anticheat on how to overload them using env vars.

walkline:
@killerwife, I've added your suggestions. However, I'm not sure about comments in configuration files. Let me know if you have any suggestions regarding this.

killerwife:
Yep, I am fine with that, I will request @cyberium and @evil-at-wow to look at this, and if they dont respond in like a week or two ill merge it.

src/shared/Config/Config.cpp
```diff
@@ -18,19 +18,24 @@
...
 
 #include <unordered_map>
 #include <string>
 #include <fstream>
+#include <cstdlib>
 
 INSTANTIATE_SINGLETON_1(Config);
 
-bool Config::SetSource(const std::string& file)
+std::optional<std::string> EnvVarForIniKey(std::string const&, std::string const&);
```
evil-at-wow, line 33: This will require an include for ```<optional>``` that's currently missing here. I'm sure it gets included indirectly somehow because the PR builds, but let's not rely on that.

src/shared/Config/Config.cpp
```diff
@@ -113,3 +136,14 @@ float Config::GetFloatDefault(const std::string& name, float def) const
     return std::stof(value);
 }
 
+std::optional<std::string> EnvVarForIniKey(std::string const& prefix, std::string const& key)
+{
+    std::string escapedKey = key;
+    std::replace(escapedKey.begin(), escapedKey.end(), '.', '_');
```
evil-at-wow, line 142: This will require an include for ```<algorithm>``` that's currently missing here.

src/shared/Config/Config.cpp
```diff
@@ -113,3 +136,14 @@ float Config::GetFloatDefault(const std::string& name, float def) const
...
+    std::string envKey = prefix + escapedKey;
+    char* val = std::getenv(envKey.c_str());
+    if (!val)
+        return std::nullopt;
```
evil-at-wow, line 146: Just a FYI: one could also simply write ```return {}``` instead of writing the ```nullopt``` explicitly.

evil-at-wow (commented):
Just some nitpicks about missing includes (cf comments), but looking OK to me.

killerwife:
Appreciate your diligence in adding the changes. I will propagate it to the other branches on next core sync.

PR #509, opened 2024-03-27
Admin controls for creature respawn rates

LunaticInAHat:
## 🍰 Pullrequest
This PR allows server admins to exercise some broad control over creature respawn rates, without having to modify the creature entries in the DB. There is an overall "rate" knob, as well as separate knobs for minimum & maximums. Together, they are useful for tuning the game experience to suit a play group's tastes:
 - Avoiding mobs rapidly respawning behind the player, while down in caves
 - Capping the amount of time that must be waited for mobs that don't spawn very frequently
 - Overall speeding up / slowing down world repopulation

### Proof
- None

### Issues
- None

### How2Test
Set non-default values in config. Find mob with known respawn time from DB. Kill mob. Wait. Observe that creature respawns at a time that respects the configuration options that were set.

### Todo / Checklist
- [X] None


LunaticInAHat:
I'm certainly not opposed to keeping this in my own fork. This was more of a "float the idea, see if people think it's useful" thing.

Corpse despawn has generally not been an issue in my testing, but I usually was trying to slow down respawn, not speed it up (so corpses would be long gone, anyways)

Also, aren't there are already knobs for controlling corpse decay times? Or are those only effective for dynguid mobs as well?

I didn't really see any fancy "dynamic" logic for respawn times, but if you point me in its direction, I would take a look at it and see if there is a better way that I can integrate what I'm trying to do with it.

cyberium:
Iam not against the idea but i feel like this code should grow a bit more to support all our current cases. But iam not even sure if it work for dynamic guids creature i think its already enough because it should be our main focus.

killerwife:
You cannot tweak corpse despawn times unless you want corpse to despawn mid someone looting it. What dynguid does is removes corpse respawn time from the respawn cycle by enabling (one alive spawn, 10 corpses). Also just to put it out there, any such global implementation will severely break bosses, world bosses, named mobs, etc, everything that isnt a common mob. (putting a multiplier on a value of lets say, 7 days, doesnt sound like fun now does it, you do not want a boss to respawn before weekly reset etc or never realistically respawn).

The official dynamic respawns basically monitor amount of people in zone, and there is a formula that decreases respawn time by up to 100% in most extreme cases. This will generally never be utilized on a small server, from what ive seen, 20 people in a zone is generally decrease of respawn time by like 25%. The value in DB is always the maximum, i.e. lets say by up to 5-10 ppl in a zone.

And then there is the fact we have like 25% world conversion to dynguid (better on tbc, worse on the other two) overall at the moment, and I am generally against altering linking and pooling in any way to force people to help with the dynguid safety conversion (i.e. removing linking and pooling)

In the end I feel like you would instead want to tweak the mobs based on their elite status, boss status etc, and I feel like at that point, what you want instead is thread safe respawn time reload and tweaking the DB. Feel free to correct me on any of my asumptions.

(and lastly, tweaking db, is free performance wise :P)

cyberium:
Performance is not an issue with that code, i think you know that already.
The OP want to increase the respawn time using this witch is understandable while we want that code cover all usages.
I agree on boss/elite issue. That may be unacceptable increasing/decreasing their respawn time. Probably even we'll get some new issue related to this.

So yes i feel like this should be redone in another way. Thank @LunaticInAHat for the proposal we greatly appreciate it and we hope to see you again with other proposals but we'll pass for this one.

LunaticInAHat:
Breaking world bosses certainly isn't good. In principle I think it would be easy to provide different respawn rates for different mob statuses -- it would just lead to a proliferation of config vars similar to what currently exists with the `Corpse.Decay.*` vars. The performance impact doesn't seem too worrisome to me; mobs don't die that frequently, and the math isn't very complicated.

Replicating the official "dynamic respawn" behavior as you describe might be an interesting thing to do, that I could look into the complexity of, but it certainly would not contribute to fixing the issues that drove me to do this in the first place. I personally haven't been having problems with mobs getting wiped out by high player density, and needing them to respawn faster. Maybe other admins have, but my server isn't that populated.

The issue I had been having is getting down into caves, and having to slog my way through the entire cave's population on the way in, and the way out, because mobs respawn after only a couple of minutes. For example, in Coldridge Valley there's a Trogg cave, and they all respawn after 3 minutes, per the DB. That may very well be the respawn rate that an official server would have, but it felt too fast to me, so I wanted to slow it down. So, my usecase was basically "I don't want anything to respawn until it's been dead for at least 10 minutes", and I added the rate & max stuff cuz I was in there doing stuff anyway.

Tweaking the DB does provide ultimate flexibility, but it's not a particularly "user-friendly" way for server admins to adjust their play group's experience. It requires a fair amount of familiarity with the DB schemas, makes it hard to take DB updates, etc.

So, I guess my thoughts are:
 - I'm perfectly happy to have this PR closed as not being something that integrates well with the overall project, and just maintain it in my own fork
 - I'm also happy to revamp this PR to distinguish between normal mobs, elite mobs, world bosses, rares, etc., and provide separate knobs for each
 - I would also be willing to look at implementing a zone-based dynamic respawn rate algorithm (possibly in conjunction with the above; I imagine the 'dynamic' part would just be a fractional multiplicative factor when calculating respawn times, so other knobs would still be needed, if we want to allow admins to slow mob respawns down)
 - I could also imagine trying a spawn-group-based dynamic respawn rate algorithm; the closer a spawn group gets to being entirely killed off, the smaller its creature's respawn timers get. I am a bit worried about the number of knobs that a scheme like this might end up needing, but still maybe worth looking at

PR #512, opened 2024-04-04
Fix broken playerbot build

Foritus:
## 🍰 Pullrequest
The build is currently broken if you try to build with BUILD_DEPRECATED_PLAYERBOT enabled (as the new playerbot is not yet ready, there isn't much option if you want bots 🙂). This fixes a missing arg when building with bots enabled.


al3xc1985:
The new playerbot system is coming. A system witch is much better and nicer than the old playerbots

Foritus:
I'm looking forward to it :) (it looks really cool!) obviously skip this if it's going to cause merge conflicts, I just thought I'd fix a build error while I was here

PR #517, opened 2024-06-17
Fix logging in/out on moving transport

insunaa:
## 🍰 Pullrequest

This PR fixes a bug that prevents you from logging in if you logged out on a moving transport.

### Proof

- None

### Issues

- None

### How2Test

- `.tele deeprun`
- Wait for a gondola
- step onto the gondola
- wait until it starts moving then log out
- log back in

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -16501,7 +16501,10 @@ bool Player::LoadFromDB(ObjectGuid guid, SqlQueryHolder* holder)
         {
             GameObjectInfo const* transportInfo = sGOStorage.LookupEntry<GameObjectInfo>(data->id);
             if (transportInfo->type == GAMEOBJECT_TYPE_TRANSPORT)
+            {
                 guid = ObjectGuid(HIGHGUID_GAMEOBJECT, data->id, transGUID);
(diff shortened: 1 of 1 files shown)


kelbren:
Tested, and works as expected on my server. 
Thanks much

insunaa:
Adjusted the PR to add @Exxenoz suggestions

Exxenoz:
The underlying problem appears to be that HIGHGUID_MO_TRANSPORT is assumed when setting t_guid at line 16472. This does not apply to every transporter. The subsequent check 'changedTransports' in AddPassenger then returns true, even though it's the same transporter.

insunaa:
Do you know if there's documentation on the difference between HIGHGUID_MO_TRANSPORTER and HIGHGUID_TRANSPORTER ?

Exxenoz:
> Do you know if there's documentation on the difference between HIGHGUID_MO_TRANSPORTER and HIGHGUID_TRANSPORTER ?

Nope, sorry. 😄 HIGHGUID_TRANSPORTER are probably elevators and possibly the Deeprun Tram (linear movements?). Presumably mainly relevant for client-side movement calculations.

PR #519, opened 2024-07-05
Implement sAuctionHouseBot.IsInitialized

mostlikely4r:
## 🍰 Pullrequest
IsInitialized() implementation for AuctionhouseBot to allow external modules (such as playerbot) to detect if GetItemData() can be used safely. 

### Proof
When AuctionHouseBot.Chance.Sell and  AuctionHouseBot.Chance.Buy are both set to 0 in the config GetItemData() will return uninitialized values. This causes issues in for example playerbot code.

### How2Test
Disable Auctionhousebot in the config with playerbots enabled.

### Todo / Checklist
- [X] Use IsInitialized in botcode.

### DIFF
--- a/src/game/AuctionHouseBot/AuctionHouseBot.cpp
+++ b/src/game/AuctionHouseBot/AuctionHouseBot.cpp
@@ -30,7 +30,7 @@
 
 INSTANTIATE_SINGLETON_1(AuctionHouseBot);
 
-AuctionHouseBot::AuctionHouseBot() : m_configFileName(_AUCTIONHOUSEBOT_CONFIG), m_houseAction(-1)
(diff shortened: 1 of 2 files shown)


mostlikely4r:
> I remember killerwife asking for comments for everything related to playerbot, for anyone who wanna have a look and help in the future to be made easy. I suggest add some comments too
> 
> Just my 2 cents

Good point. I added a comment for the new method:
 // Following method was added for playerbot interaction so GetItemData is only used when AuctionHouseBot is enabled in config.

al3xc1985:
Perfect

mostlikely4r:
No longer needed with upcoming bot changes.

PR #533, opened 2024-07-30
Spell: 42793 corrected

Grz3s:
double credit fixed


Grz3s:
Ref:
https://github.com/cmangos/wotlk-db/issues/1321

PR #536, opened 2024-08-09
StaticFlags: Implement CANNOT_TURN

insunaa:
## 🍰 Pullrequest

This PR implements CreatureStaticFlags3::CANNOT_TURN

### DIFF
--- a/src/game/MotionGenerators/TargetedMovementGenerator.cpp
+++ b/src/game/MotionGenerators/TargetedMovementGenerator.cpp
@@ -934,8 +934,10 @@ bool FollowMovementGenerator::_getOrientation(Unit& owner, float& o) const
 {
     if (!i_target.isValid())
         return false;
-
-    o = (i_faceTarget ? owner.GetAngle(i_target.getTarget()) : i_target->GetOrientation());
+    if (owner.CannotTurn())
+        o = owner.GetOrientation();
+    else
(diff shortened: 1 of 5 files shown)


src/game/MotionGenerators/TargetedMovementGenerator.cpp
```diff
@@ -934,8 +934,10 @@ bool FollowMovementGenerator::_getOrientation(Unit& owner, float& o) const
 {
     if (!i_target.isValid())
         return false;
-
-    o = (i_faceTarget ? owner.GetAngle(i_target.getTarget()) : i_target->GetOrientation());
+    if (owner.CannotTurn())
+        o = owner.GetOrientation();
```
insunaa, line 938: maybe
```cpp
return false;
```
?

killerwife: This is prolly not necessary. Likely client will handle the "not turning", i.e. when npc has smth in crosshairs, then client wont turn npc to it

insunaa: Yes, but I've also seen cases with Kologarn for example where he's not supposed to turn, but then when he dies and plays his death animation his corpse falls sideways, or he turns for a millisecond while doing an attack and then snaps back forward, so I added this to be safer, I guess.

PR #541, opened 2024-08-09
StaticFlags: Implement NO_AUTOMATIC_REGEN

insunaa:
## 🍰 Pullrequest

This PR implements CreatureStaticFlags::NO_AUTOMATIC_REGEN

### DIFF
--- a/src/game/Entities/Creature.cpp
+++ b/src/game/Entities/Creature.cpp
@@ -597,7 +597,8 @@ bool Creature::UpdateEntry(uint32 Entry, const CreatureData* data /*=nullptr*/, 
     if (IsWorldBoss())
         ApplySpellImmune(nullptr, IMMUNITY_STATE, SPELL_AURA_MOD_TOTAL_STAT_PERCENTAGE, true);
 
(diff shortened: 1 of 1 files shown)


src/game/Entities/Creature.cpp
```diff
@@ -600,6 +600,9 @@ bool Creature::UpdateEntry(uint32 Entry, const CreatureData* data /*=nullptr*/,
     if ((GetCreatureInfo()->RegenerateStats & (REGEN_FLAG_POWER_IN_COMBAT | REGEN_FLAG_POWER)) == (REGEN_FLAG_POWER_IN_COMBAT | REGEN_FLAG_POWER))
         SetFlag(UNIT_FIELD_FLAGS_2, UNIT_FLAG2_REGENERATE_POWER);
 
+    if (m_settings.HasFlag(CreatureStaticFlags::NO_AUTOMATIC_REGEN))
```
killerwife, line 603: Hate to be a dick about this one, but if you set it and reset it, it will be internally moved as if it updated. Incorporate it into the set above, likely tiered if will make it more readable.

insunaa: changed it, but not sure if the flag could possibly be present even without the explicit setting above, so I didn't delete the "RemoveFlag" part (yet)

PR #549, opened 2024-08-16
SafeLoc: Implement Westfall GY as fallback graveyard

insunaa:
## 🍰 Pullrequest

This PR ensures that if no GY can be found for a given character, Westfall is chosen.


Redbu11dev:
maybe is not very good to have westfall as fallback for both teams? maybe make it westfall only for alliance, and something else for horde?

insunaa:
Unfortunately that is how it is on the reference servers. Both factions always get sent to Westfall when no valid gy can be determined

insunaa:
ok I've added crossroads for horde, pending confirmation that this was how it actually worked in WotLK. From my memory it always used to be Sentinel Hill, and my guess was that Crossroads was added for Horde in a later expansion, but because I don't currently have confirmation either way I added it.

If Team is not Horde it will always to go Sentinel Hill, but if team is definitely Horde then it's going to Crossroads

Redbu11dev:
nice 👍

PR #555, opened 2024-08-31
Timer: Add a timer for code-benchmarking with RAII

insunaa:
## 🍰 Pullrequest

This PR adds a timer to benchmark code performance.

To use it simply create a new ChronoTimeTracker object and let it go out of scope, or delete it. Example:

```cpp
{
    ChronoTimeTracker bench;
    // some code you want to benchmark
}
```
or
```cpp
ChronoTimeTracker* bench = new ChronoTimeTracker();
// some code you want to benchmark
delete bench;
```

The time can also be polled at runtime with
```cpp
ChronoTimeTracker bench;
while(1)
{
    sLog.outDebug("%ld", bench.elapsedSeconds().count());
}
```

Please note that measurements taken like this come with variance and should only be trusted in aggregate

### DIFF
--- a/src/shared/Util/CodeBench.cpp
+++ b/src/shared/Util/CodeBench.cpp
@@ -1,0 +1,13 @@
+#include "CodeBench.h"
+#include "Log/Log.h"
+
+ChronoTimeTracker::~ChronoTimeTracker()
+{
+    std::chrono::nanoseconds elapsed = elapsedNanos();
+    sLog.outError("%s: time elapsed: %ldns, %ldµs, %ldms, %lds",
+        m_name.c_str(),
+        nanos(elapsed).count(),
+        micros(elapsed).count(),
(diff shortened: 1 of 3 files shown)


killerwife:
This is why we have grafana integration. Adding two lines of code in places can be done on anyones local. Not sure this is needed at all. (as a matter of fact i used to do it regularly, without the need for raii)

insunaa:
yeah, the grafana-based metrics are really neat for long-term review of things, but they don't help much while actively working on some code. with this I can just plop in an object, compile run and see immediately in console if what I did was garbage or decent. With the regular metrics it is powerful when observing a running server over time

killerwife:
I have warmed up to the idea, under the condition you put it in its own separate header file, not where you put it currently. And put the destructor implementation in a cpp file so log.h doesnt have to be in the header.

PR #571, opened 2024-11-12
Fix CodeBench guard macro

insunaa:
## 🍰 Pullrequest

Old Guard Macro already exists in timer code, so can't be re-used.

Imo we should just switch to
```cpp
#pragma once
```
it is supported by every compiler and build system that is even remotely relevant.

Also fixes ambiguous include statement.

### DIFF
--- a/src/shared/Util/CodeBench.cpp
+++ b/src/shared/Util/CodeBench.cpp
@@ -16,7 +16,7 @@
  * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
  */
 
-#include "CodeBench.h"
+#include "Util/CodeBench.h"
 #include "Log/Log.h"
 
 ChronoTimeTracker::~ChronoTimeTracker()
--- a/src/shared/Util/CodeBench.h
+++ b/src/shared/Util/CodeBench.h
@@ -16,8 +16,8 @@
  * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
  */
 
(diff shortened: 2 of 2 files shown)


killerwife:
Oops

evil-at-wow:
Related to the suggestion to switch to ```#pragma once```, note that some compilers recommend against using it. For example, [the GCC devs say, in their latest version](https://gcc.gnu.org/onlinedocs/gcc-14.2.0/cpp/Alternatives-to-Wrapper-_0023ifndef.html):

> (...) Neither one is as portable as a wrapper ‘#ifndef’ and we recommend you do not use them in new programs, ...

As far as I understand, their main reasons seems to be that, generally speaking, figuring out if two files are the same file on a file system is not a trivial task (think symlinks and especially hardlinks). I remember GCC fixing their implementation of ```#pragma once``` because of such issues in the past.

I also wonder what happens if you copy such a file as part of a build. I know the Qt project copies - or at least used to copy - header files to other folders when preparing the build. In such cases, clearly those copies would be different files then, so ```#pragma once``` would still allow to include them both while a simple include guard would not.

So I'm with GCC and in favor of a simply yet effective include guard (which your editor should insert for you anyway to avoid copy/paste mistakes :stuck_out_tongue:).

insunaa:
Our build process doesn't copy files. GCC officially supports `#pragma once` whether they like it or not.
CMaNGOS is not made in Qt, and if they copy around files for building they're playing with fire regardless.

I'm not vehemently opposed to include guards, but I think they're archaic and it's up to the compiler devs and build system devs to ensure that their garbage works, imo

PR #574, opened 2024-11-12
Battleground: Fix crash when no players are left in a battleground.

mostlikely4r:
## 🍰 Pullrequest
This is a small crashfix which occurs when a battleground ends up empty.

### Proof
The code looks pretty clear m_bg is set to nullptr and the next BattleGroundMap::Update does m_bg->
While I'm unsure if there should have been some mechanic to prevent the update after the m_bg is cleared it clearly does not work.
It's very trivial to reproduce and test though.

### How2Test
-Join a battleground (".debug bg" to join alone).
-Hearthstone out and the server crashes.

### DIFF
--- a/src/game/Maps/Map.cpp
+++ b/src/game/Maps/Map.cpp
@@ -2112,6 +2112,9 @@ void BattleGroundMap::Update(const uint32& diff)
 {
     Map::Update(diff);
 
+    if (!m_bg)
+        return;
+
     if (!m_bg->GetPlayersSize())
     {
         // BG is empty


killerwife:
In what place in code is it set to nullptr if i may ask. Because this might change with the BG rewrite.

mostlikely4r:
> In what place in code is it set to nullptr if i may ask. Because this might change with the BG rewrite.

https://github.com/cmangos/mangos-wotlk/blob/6a6de7bc8cf4d96684f200d2d67e645a42fea8ee/src/game/Maps/Map.cpp#L2133
So the first BattleGroundMap::Update sets m_bg to nullptr and the second crashes because it is nullptr.

mostlikely4r:
I can no longer reproduce this right now. I've seen it happen in tbc but not consistently so I'll investigate further first.

kelbren:
Using cmangos wotlk and indeed the server crashes if I enter a battleground alone and hearth out. every time.
the proposed change seems to resolve the crashing.

mostlikely4r:
> Using cmangos wotlk and indeed the server crashes if I enter a battleground alone and hearth out. every time. the proposed change seems to resolve the crashing.

Are you using latest commits? I'll have to check again on my end.

PR #581, opened 2024-12-22
Boost: Add 1.87 support

insunaa:
## 🍰 Pullrequest

This PR replaces the deprecated io_service with io_context, which is functionally identical.


src/shared/Network/AsyncListener.hpp
```diff
@@ -31,7 +31,7 @@ namespace MaNGOS
     {
         public:
             // constructor for accepting connection from client
-            AsyncListener(boost::asio::io_service& io_service, std::string const& bindIp, unsigned short port) : m_service(io_service), m_acceptor(io_service, boost::asio::ip::tcp::endpoint(boost::asio::ip::address::from_string(bindIp), port))
+            AsyncListener(boost::asio::io_context& io_service, std::string const& bindIp, unsigned short port) : m_service(io_service), m_acceptor(io_service, boost::asio::ip::tcp::endpoint(boost::asio::ip::address::from_string(bindIp), port))
```
evil-at-wow, line 34: The only (nitpick) comment I have is that maybe we should rename the parameter name as well, next to its type.

evil-at-wow (approved):

src/shared/Network/AsyncListener.hpp
killerwife: What he said.

src/shared/Network/AsyncSocket.hpp
```diff
@@ -74,7 +74,7 @@ namespace MaNGOS
     };
 
     template <typename SocketType>
-    MaNGOS::AsyncSocket<SocketType>::AsyncSocket(boost::asio::io_service& io_service) : m_socket(io_service), m_address("0.0.0.0"),
+    MaNGOS::AsyncSocket<SocketType>::AsyncSocket(boost::asio::io_context& io_service) : m_socket(io_service), m_address("0.0.0.0"),
```
insunaa, line 77: looks like I missed some

insunaa:
Added more work from Niam and myself. Requires someone who actually uses metrics to check, tho.

killerwife:
Metrics are my problem so we need not worry about it right now. PR is ready as far as I am concerned. Rebase and squash at your leisure.

killerwife (approved):

insunaa:
@killerwife squashed the commits but I kept the cleanup commit for Socket.cpp separate. should I merge it together with this PR or do it separately?

killerwife:
Yep thats the right call to keep it separate

PR #582, opened 2025-01-02
Mounts: Fix issues with vehicles and flying mounts

insunaa:
## 🍰 Pullrequest

This PR fixes a crash in Vehicle code that is caused by the vehicleInfo being destroyed before UnBoard is called, with UnBoard necessarily referencing vehicleInfo, which is its parent.

This PR also ensures that flying mounts have their speed properly updated. Don't ask my why it wasn't before.

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -2064,7 +2064,8 @@ bool Player::ResolvePendingUnmount()
         ResummonPetTemporaryUnSummonedIfAny();
 
     UpdateSpeed(MOVE_RUN, true); // update speed
-    UpdateSpeed(MOVE_FLIGHT, true);
+    if (m_pendingMountAuraFlying)
+        UpdateSpeed(MOVE_FLIGHT, true);
 
     return true;
 }
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -2064,6 +2064,7 @@ bool Player::ResolvePendingUnmount()
         ResummonPetTemporaryUnSummonedIfAny();
 
     UpdateSpeed(MOVE_RUN, true); // update speed
+    UpdateSpeed(MOVE_FLIGHT, true);
 
     return true;
 }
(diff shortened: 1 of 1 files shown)


src/game/Entities/Player.cpp
```diff
@@ -2040,6 +2040,8 @@ bool Player::ResolvePendingMount()
     }
 
     UpdateSpeed(MOVE_RUN, true); // update speed
+    if (m_pendingMountAuraFlying)
+        UpdateSpeed(MOVE_FLIGHT, true);
```
killerwife, line 2044: It worked before because before it was done in place in aura. Now we are in an async environment and need to do it when client confirms the collision height check. Has to be done on unmount too however

src/game/Entities/Vehicle.cpp
```diff
@@ -197,6 +197,14 @@ VehicleInfo::VehicleInfo(Unit* owner, VehicleEntry const* vehicleEntry, uint32 o
 
 VehicleInfo::~VehicleInfo()
 {
+    std::vector<Unit*> passengers;
```
killerwife, line 200: I want to see a stacktrace for this crash if possible to see when the destructor is called and not unboarded yet.

src/game/Entities/Player.cpp
```diff
@@ -2062,6 +2064,7 @@ bool Player::ResolvePendingUnmount()
         ResummonPetTemporaryUnSummonedIfAny();
 
     UpdateSpeed(MOVE_RUN, true); // update speed
+    UpdateSpeed(MOVE_FLIGHT, true);
```
insunaa, line 2067: I've added an unconditional flight speed update on dismount, because I'm unsure whether there can be a pending mount with a ground-mount before the pending unmount is executed, which would override the `m_pendingMountAuraFlying` variable

killerwife: Make it conditional, for two reasons, its a heavy function and previously it was conditional on having the aura as well.

PR #583, opened 2025-01-04
Vehicles: Additional cleanup

insunaa:
## 🍰 Pullrequest

This PR adds a Vehicle-Cleanup call in `CleanupsBeforeDelete`

### DIFF
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -10631,8 +10631,11 @@ void Unit::CleanupsBeforeDelete()
         ClearComboPointHolders();
         RemoveAllAuras(AURA_REMOVE_BY_DELETE);
     }
-    if (IsVehicle())
+    if (m_vehicleInfo)
+    {
         m_vehicleInfo->Cleanup();
+        m_vehicleInfo = nullptr;
+    }
     WorldObject::CleanupsBeforeDelete();
 }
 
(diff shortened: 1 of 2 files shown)


insunaa:
Keeping this PR open for a bit longer to see if I find any more preventable crashes

src/game/Entities/Unit.cpp
```diff
@@ -10631,8 +10631,11 @@ void Unit::CleanupsBeforeDelete()
         ClearComboPointHolders();
         RemoveAllAuras(AURA_REMOVE_BY_DELETE);
     }
-    if (IsVehicle())
+    if (m_vehicleInfo)
+    {
         m_vehicleInfo->Cleanup();
+        m_vehicleInfo = nullptr;
```
killerwife, line 10637: I believe the nullptr is unnecessary. It will happen when unit is deleted

insunaa: I thought so too, and then I had a crash
image

killerwife: Well, I mean sure, that fails due to MANGOS_ASSERT(m_passengers.empty()); but I fail to see how calling destructor (by setting the ptr to nullptr) fixes that.

insunaa: Because setting it to nullptr early means that `IsVehicle()` or `if (m_vehicleInfo)` returns false, both of which are used in other aspects of the cleanup.

I haven't yet been able to provoke this same crash again, the server has been running stably for ~2 hours now with no crashes since I added the `= nullptr;` and I'll keep it running some more to see if maybe the crash re-occurs. If it doesn't then I'll just assume setting it to `nullptr` fixed the issue. If it does I'll have to investigate some more. In any case this crash occurred as shown in the crashlog and I couldn't find any other reason for why passengers might not be empty when the crash occurred

PR #584, opened 2025-01-12
Deprecated Playerbots: Fix CreatureTypeFlags

insunaa:
## 🍰 Pullrequest

This PR fixes the deprecated playerbots compile process to work with the new CreatureTypeFlags

### Issues

- fixes https://github.com/cmangos/issues/issues/3828


evil-at-wow:
Clearly I need to check PRs more often, especially in the non-TBC cores. I fixed this a few days ago as a side effect of merging my TBC 'fix old playerbots' change and noticing it needed something more in the WoTLK branch (see cmangos/mangos-wotlk@cebfb50a42cd7499cf05fb2ce6374f57cbfee08c). If I had seen this before, I'd have merged this first...

PR #589, opened 2025-01-30
Playerbot: linux compile fix for GCC and CLang

mostlikely4r:
## 🍰 Pullrequest
This is a small include to fix a linux compile error specifically for GCC 14.2.1 / CLang 18.1.8

### Proof
In the playerbot specific code sMapStore.LookupEntry is used which required this include.

### DIFF
--- a/src/game/MotionGenerators/PathFinder.cpp
+++ b/src/game/MotionGenerators/PathFinder.cpp
@@ -25,6 +25,9 @@
 #include "Entities/Transports.h"
 #include <Detour/Include/DetourCommon.h>
 #include <Detour/Include/DetourMath.h>
+#ifdef ENABLE_PLAYERBOTS
+#include "Server/DBCStores.h"
+#endif
 
 #ifdef BUILD_METRICS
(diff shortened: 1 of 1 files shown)


evil-at-wow:
See cmangos/mangos-wotlk@99934807e00e67925d3478cfbaf4acfdaac94e79, but it seems I was a bit too fast/sloppy. After I 'merged' another unrelated PR with merge commits by cherry-picking the actual commit we want for that PR, I accidentally did the same thing for yours. So now GitHub doesn't see your PR is merged :cry:

mostlikely4r:
> See [9993480](https://github.com/cmangos/mangos-wotlk/commit/99934807e00e67925d3478cfbaf4acfdaac94e79), but it seems I was a bit too fast/sloppy. After I 'merged' another unrelated PR with merge commits by cherry-picking the actual commit we want for that PR, I accidentally did the same thing for yours. So now GitHub doesn't see your PR is merged 😢

As long as the code change is inside the master I don't mind. Thanks for informing me.

PR #590, opened 2025-02-17
fix: renamed playerbots build-flag in order to fix playerbot builds

AlexPanacea:
## 🍰 Pullrequest
I fixed the way the nightly build is building the playerbot-related binaries.
Because of new way the playerbot module is used/added a small naming changed cause the buildflag to be misread.

I've tested it locally and playerbot seems to work like normal the nightly build also passed last night.

Before:
image

After:
image
Notice how the playerbot related builds are significantly larger now.


AlexPanacea:
How can I help out with getting this small fix merged?

al3xc1985:
well, my advice is to just wait. Killwewife has a lot to do.
Maybe @evil-at-wow can help on this one

evil-at-wow:
That's an interesting typo indeed, good find. That's probably a remnant from the old and now deprecated playerbots code before that was renamed and the new playerbot code was added. Thanks!

There was a merge commit in your history/PR here, so I've taken the liberty to remove that and just use your actual fix commit because we generally don't use merge commits like that (and thus closing this PR likely makes GitHub think there are still unmerged commits, which is technically true but practically speaking we're good). So for future reference: when you update your local branch, you can just rebase it on the updated master branch, rather than merge the updated master branch to your PR branch, and then (force) push your local branch to update the PR.

And sorry for the delays, but there are not a lot of active devs around here so things tend to take a (good) while. Personally, I'm also mostly focused on the TBC core, so I admit I rarely check the Wrath/Classic cores...

AlexPanacea:
Thank you for adding my fix to the repository! In the future I'll try to avoid having merge commits into my PRs.
No worries about the delays, I know there is a lot of ongoing work and only so many hands to handle it all. (:

PR #601, opened 2025-07-30
Spell: 54432

Grz3s:
scripted


src/game/AI/ScriptDevAI/scripts/northrend/grizzly_hills.cpp
```diff
@@ -753,6 +753,34 @@ struct AttentionSpan : public AuraScript
...
+
+    void OnEffectExecute(Spell* spell, SpellEffectIndex effIdx) const override
+    {
+        if (effIdx != EFFECT_INDEX_0)
+            return;
+
+        Unit* caster = spell->GetCaster();
+        Unit* target = spell->GetUnitTarget();
+
+        Player* player = (Player*)target;
```
killerwife, line 772: Do this AFTER the check if target is player

And change the C style cast to static_cast<Player*>(target);

PR #605, opened 2025-08-28
Visibility: Fix crash when entry of large creature is changed

insunaa:
## 🍰 Pullrequest

This PR fixes a crash caused by changing the Entry of a creature that is marked as a large object to the entry of a creature that is not marked as a large object, upon the despawning of said new object.

Edit: The crash as it was reported actually did have the NPC that InitEntry() was used for also as a Large Creature, but because the Entry was no longer the same, `Map::Remove()` couldn't delete the creature from the `m_largeObjects` set, causing later map updates to access this already removed object.

### How2Test

- Learn skinning to at least 310
- Teleport to Hellfire Ramparts
- Start the Nazruden the Herald encounter
- Kill Nazruden
- Kill Nazan
- Skin Nazan
- Wait for crash

### Todo / Checklist

- [x] None

### DIFF
--- a/src/game/Entities/Creature.cpp
+++ b/src/game/Entities/Creature.cpp
@@ -354,6 +354,10 @@ bool Creature::InitEntry(uint32 Entry, CreatureData const* data /*=nullptr*/, Ga
         }
     }
 
+    // Remove from large objects list if the new entry is not also a large object
+    if (!cinfo->IsLargeOrBiggerCreature())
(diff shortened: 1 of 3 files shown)


insunaa:
OK this commit fixes the crash and re-adds the object to the largeobject list if the new entry is also a largeobject.

insunaa:
yup, sorry. your fix worked

PR #606, opened 2025-08-31
Tweak conditional compilation and linking for MinGW compatibility

Kashemir001:
## 🍰 Pullrequest

Changes two occasions of conditional statements for compiler and linker from being MSVC only to Win32 in general, which fixes project building when using MinGW toolchain. Otherwise build process fails for `game` and `mangosd` targets with undefined references.

### Issues

- None

### How2Test

Build using MinGW-W64 v13, GCC 15.2, UCRT or MSVCRT runtimes (tested with both), CMake 3.31, Boost 1.87. MinGW binary releases can be taken from https://github.com/niXman/mingw-builds-binaries. Tested for all three mainline CMaNGOS cores.

### DIFF
--- a/src/game/Maps/Map.cpp
+++ b/src/game/Maps/Map.cpp
@@ -797,7 +797,7 @@ void Map::Update(const uint32& t_diff)
 
     m_curTime = time(nullptr);
 
-#ifdef _MSC_VER
+#ifdef _WIN32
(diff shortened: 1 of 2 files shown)


killerwife:
I have no problem with this being merged apart from my comment, but take note that it wont get merged until after 18th since i am away.

PR #634, opened 2025-12-23
Fix compiler warnings about enums and arithmetics

ssj17vegeta:
## 🍰 Pullrequest
Force casting enums when used for arithmetic operations into integer or float values (said types will depend on their intended usage, i.e the signature of the method that takes them).

### DIFF
--- a/src/game/BattleGround/BattleGround.cpp
+++ b/src/game/BattleGround/BattleGround.cpp
@@ -382,7 +382,7 @@ void BattleGround::Update(uint32 diff)
     if (IsArena() && !m_arenaBuffSpawned)
     {
         // 60 seconds after start the buffobjects in arena should get spawned
-        if (m_startTime > uint32(m_startDelayTimes[BG_STARTING_EVENT_FIRST] + ARENA_SPAWN_BUFF_OBJECTS))
+        if (m_startTime > uint32(static_cast<uint32>(m_startDelayTimes[BG_STARTING_EVENT_FIRST]) + static_cast<uint32>(ARENA_SPAWN_BUFF_OBJECTS)))
         {
             SpawnEvent(ARENA_BUFF_EVENT, 0, true);
             m_arenaBuffSpawned = true;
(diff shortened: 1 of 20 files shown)


src/game/BattleGround/BattleGround.cpp
```diff
@@ -382,7 +382,7 @@ void BattleGround::Update(uint32 diff)
     if (IsArena() && !m_arenaBuffSpawned)
     {
         // 60 seconds after start the buffobjects in arena should get spawned
-        if (m_startTime > uint32(m_startDelayTimes[BG_STARTING_EVENT_FIRST] + ARENA_SPAWN_BUFF_OBJECTS))
+        if (m_startTime > uint32(static_cast<uint32>(m_startDelayTimes[BG_STARTING_EVENT_FIRST]) + static_cast<uint32>(ARENA_SPAWN_BUFF_OBJECTS)))
```
evil-at-wow: The C-style cast ```uint32(...)``` can be removed because they are unnecessary if we already do a ```static_cast```. Similar below.

src/game/BattleGround/BattleGroundAV.h
```diff
@@ -670,7 +670,7 @@ class BattleGroundAV : public BattleGround
         void ChangeMineOwner(AVMineIds mineId, PvpTeamIndex newOwnerTeamIdx);
 
         // World state helpers
-        uint8 GetWorldStateType(uint8 state, PvpTeamIndex teamIdx) const { return teamIdx * BG_AV_MAX_STATES + state; }
+        uint8 GetWorldStateType(uint8 state, PvpTeamIndex teamIdx) const { return static_cast<uint16>(teamIdx) * static_cast<uint16>(BG_AV_MAX_STATES) + state; }
```
evil-at-wow: This could potentially cast to ```uint8``` given the function's return type? The potential for overflow (based on types) is already there anyway.

ssj17vegeta:
@evil-at-wow thanks for reviewing this PR. I added a second commit that addresses your points ;)

PR #636, opened 2025-12-23
Fix server crash on player login, part 2

ssj17vegeta:
## 🍰 Pullrequest
Second part of my original pull request that addressed the same problem which was dealt with in another commit, but only partially. It still crashes without this modification (inspired by @killerwife 's solution).

### Issues
- relates https://github.com/cmangos/issues/issues/4017

### How2Test
(Not all) characters crash on login.

### DIFF
--- a/src/game/Achievements/AchievementMgr.cpp
+++ b/src/game/Achievements/AchievementMgr.cpp
@@ -772,7 +772,7 @@ void AchievementMgr::SendCriteriaUpdate(uint32 id, CriteriaProgress const* progr
     WorldPacket data(SMSG_CRITERIA_UPDATE, 8 + 4 + 8);
     data << uint32(id);
 
-    TimePoint now = GetPlayer()->GetMap()->GetCurrentClockTime();
+    TimePoint now = sWorld.GetCurrentClockTime();
     // the counter is packed like a packed Guid
     data.appendPackGUID(progress->counter);
 


killerwife:
Okay can you give me the crash back trace for this one? Because yes, it will fix the issue, HOWEVER

It clashes with the aspirational threading model, where world is async from map.

SendCriteriaUpdate is used from stuff like StartAchievementCriteria which is clearly Map Thread, and in all those cases, we must at all cost use Map time fetch.

Achievements arent my forte so I am not simply able to gleam the code path necessary for it crashing in this particular case, which will still be I assume on load or some other shenanigans. However should in the future I manage to asynchronize map and world like I would love to, this code would suddenly be a landmine (currently its the same timestamp)

GPinkie:
[b07db02d_mangosd.exe_[28-3_15-34-7].txt](https://github.com/user-attachments/files/26617792/b07db02d_mangosd.exe_.28-3_15-34-7.txt)

Fixed this achievement crash i was getting.

killerwife:
Okay so, I was kinda correct. We do 

m_achievementMgr.CheckAllAchievementCriteria

after adding player to map.

Hence all checks until that point in loading process, need to be skipped. Should more of these pop up, i will fix them as I go. Ty for the help on this.

Linked the commit in the pr

PR #642, opened 2026-01-31
Fix client crash when a session is reused

slnz00:
## 🍰 Pullrequest
The client crashes when a chat message is sent after it was previously closed without properly logging out.

The game crashes with an access violation at `0x00530475` caused by a null pointer dereference. The crash occurs because the server does not send the tutorial data flags when a session is reused after a client reconnects. The pointer to the tutorial flags is stored in a global variable at `0x00BD1AFC`, which remains null if the client does not receive this data from the server, leading to a null pointer exception when the client checks whether the chat related tutorial flag is set.

### Proof

- None

### Issues
- None

### How2Test
- Start the game and log in
- Send a chat message -> all good, works fine
- Close the game with Alt+F4
- Restart and log in again using the same account
- Send another message -> client crashes

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Server/WorldSocket.cpp
+++ b/src/game/Server/WorldSocket.cpp
@@ -571,8 +571,6 @@ bool WorldSocket::HandleAuthSession(WorldPacket& recvPacket)
 
             std::unique_ptr<SessionAnticheatInterface> anticheat = sAnticheatLib->NewSession(session, K);
 
-            session->SendAuthOk();
-
             // when false, the client sent invalid addon data.  kick!
             WorldPacket addonPacket; // yes its copypasted atm cos of reconnect
             if (!anticheat->ReadAddonInfo(const_cast<WorldPacket*>(&recvPacket), addonPacket))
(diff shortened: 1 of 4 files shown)


src/game/Server/WorldSocket.cpp
```diff
@@ -571,8 +571,6 @@ bool WorldSocket::HandleAuthSession(WorldPacket& recvPacket)
 
             std::unique_ptr<SessionAnticheatInterface> anticheat = sAnticheatLib->NewSession(session, K);
 
-            session->SendAuthOk();
```
slnz00: Sending `AuthOk` here silently fails because the reused session does not have an `m_socket` member yet. The new socket is stored in `m_requestSocket`, and `m_socket` is not set until the `WORLD_SESSION_STATE_CREATED` state is handled. ([WorldSession.cpp:496](https://github.com/cmangos/mangos-wotlk/pull/642/changes#diff-ef89bb660d83c50859b303b6187cc4191de702af86f3a6db40c0b0db13e4cb5a)). The anticheat related addon check should also be reviewed. Sending the packet works because it uses the socket defined on the `WorldSocket` instance. However if this is meant to run after the player has already logged in, it should be moved as well, since the client has not received the `AuthOk` response at this point.

killerwife:
Appreciate the contribution. Sorry for the time it took to merge.

PR #647, opened 2026-04-11
Fix cooldowns

### LINKED ISSUE #3418: 🐛 [Bug Report] [WOTLK] Ability cooldown (shield wall)
### Bug Details

Im pretty sure this was working fine, but this time I somehow managed to click it while logging in so it was moving the ability from the actionbar and then I just clicked it back to its place, and the cooldown showed this:

EDIT: This actually seems to happen everytime.

image

### Steps to Reproduce
EDIT: 
1. Use the ability (it will show normal cooldown, however it wont work after that as the actual cooldown is 12h)
2. Logout
3. Login (now will show 12h)

### Expected behavior

Should have normal cooldown.

### Suggested Workaround

_No response_

### Crash Log

_No response_

### Core SHA1 Commit Hash

*

### Database SHA1 Commit Hash

*

### Operating System

Win10

### Client Version

4.3.5a (Wrath of the Lich King)

### LINKED ISSUE #3877: 🐛 [Bug Report] Shield Wall Cooldown
### Bug Details

Warrior ability Shield Wall has a 300,000 second cooldown (83 hour)

### Steps to Reproduce

1. Make warrior
2. Levelup ~30
3. add any shield and 1h weapon
4. learn defensive stance (spell 71)
5. Push button.

### Expected behavior

The cooldown should be 5 minutes

### Suggested Workaround

update the spell entry (871)'s RecoveryTime to 300 (5 min)

### Crash Log

_No response_

### Core SHA1 Commit Hash

0793bb8578b31c919be3f1fab226c49d20d3ca42

### Database SHA1 Commit Hash

efbd8fecb3d212f98161fafe14756829be749491

### Operating System

Ubuntu

### Client Version

3.3.5a (Wrath of the Lich King)

### LINKED ISSUE #4021: 🐛 [Bug Report] Blind spell (id 2094) cooldown messed up
### Bug Details

The cooldown for the "Blind" rogue skill is bugged.

- When I cast "blind", I get a starting cooldown of two minutes = OK ✅.
- If, during the cooldown, I try to cast blind again, it says the skill is under cooldown = OK ✅.
- I wait for 2 minutes (the skill's official cooldown).
- "Skill is not ready to launch"

That's where it gets _very_ interesting : 
- Typing ".cooldown list" shows "Found 0 cooldown".
- If I do a /reloadui and re-type ".cooldown list", I see something : Image
- If I type ".cooldown list" DURING the first 2 minutes (where the "real" cooldown happens, I see this : Image
- If I log out and log back in, it gets even more intense, the client finally has the information that the skill is under cooldown (both in skill book and action bars), it indicates a 12 hours cooldown lol
 Image

- But that cooldown is huge AND doesn't match with .cooldown list which still indicates "Spell(2094) have , CatRecTime(41m 49s 856ms) cd"
- In that case (no "RecTime" but "CatRecTime" only), the command  ".cooldown clear" does NOT reset the skills' cooldown.

There could be other spells impacted but I haven't stumbled upon them.

Finally, here's what's in DB (since I logged out it saved) in the "character_spell_cooldown" table : 

Image

The "SpellExpireTime" field seems correct (about 15 minutes ago, the time it took me to write this issue), and the "CategoryExpireTime" is 12 hours from now.

### Steps to Reproduce

1. Log in with a horde rogue in Orgrimmar.
2. Cast Blind on Gamon.
3. Recasting blind during its "official" cooldown (2 minutes) helps reproducing the bug.
4. Wait for 2 minutes (the spell's official cooldown) and try recasting it.

### Expected behavior

The cooldown should behave normally.

### Suggested Workaround

_No response_

### Crash Log

_No response_

### Core SHA1 Commit Hash

b583c031d84755119af3bd49c3e9505d130e7f7c

### Database SHA1 Commit Hash

c2d2024c8d3e20ab31f6bb036ddeed154eeac238

### Operating System

Fedora 43

### Client Version

3.3.5a (Wrath of the Lich King)

ssj17vegeta:
## 🍰 Pullrequest
Fixes 2 cooldown problems where spells like Blind with CategoryRecoveryTime were stored with an enormous (12h+) value which was caused by an unsigned integer underflow. My original issue gave a detailed view of how the bug occurs and what happens in the database when the player logs off.

Aside from fixing the underflow, my fix takes care of the useless persisting of expired cooldowns on logout / .save (which was a secondary bug linked to the unsigned integer underflow one).

### Issues
- fixes [#4021](https://github.com/cmangos/issues/issues/4021)
- fixes [#3877](https://github.com/cmangos/issues/issues/3877)
- fixes [#3418](https://github.com/cmangos/issues/issues/3418)

### How2Test
- Horde rogue, target Gamon
- Cast Blind
- While on cooldown, try to recast blind
- When the 2 mins cooldown is finished, you should be able to cast Blind again :)
- Logout and back on, the cooldown is accurate and not 12+ hours.

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -4198,6 +4198,7 @@ void Player::_SaveSpellCooldowns()
 
     static SqlStatementID insertSpellCooldown;
 
+    uint64 nowUnix = static_cast<uint64>(Clock::to_time_t(GetMap()->GetCurrentClockTime()));
     for (auto& cdItr : m_cooldownMap)
     {
         auto& cdData = cdItr.second;
@@ -4209,6 +4210,10 @@ void Player::_SaveSpellCooldowns()
             cdData->GetCatCDExpireTime(cTime);
             uint64 spellExpireTime = uint64(Clock::to_time_t(sTime));
             uint64 catExpireTime = uint64(Clock::to_time_t(cTime));
+
+            // Skip entries where both cooldowns have already expired - no point persisting them
+            if (spellExpireTime <= nowUnix && catExpireTime <= nowUnix)
+                continue;
 
             stmt = CharacterDatabase.CreateStatement(insertSpellCooldown, "INSERT INTO character_spell_cooldown (guid, SpellId, SpellExpireTime, Category, CategoryExpireTime, ItemId) VALUES( ?, ?, ?, ?, ?, ?)");
             stmt.addUInt32(GetGUIDLow());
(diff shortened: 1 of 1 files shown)


src/game/Entities/Player.cpp
```diff
@@ -4210,6 +4211,10 @@ void Player::_SaveSpellCooldowns()
             uint64 spellExpireTime = uint64(Clock::to_time_t(sTime));
             uint64 catExpireTime = uint64(Clock::to_time_t(cTime));
 
+            // Skip entries where both cooldowns have already expired - no point persisting them
```
killerwife, line 4212: Silly q, why not do the check against the TimePoint cTime above, then you do not have to resort to the nowUnix typecast fugly and rely on type checking for a time type :D

Under the hood its still a long long in the end.

killerwife (approved):

src/game/Entities/Player.cpp
ssj17vegeta: That's not silly at all, that was me being tired... Gotta stop pulling all-nighters on code lol

So yeah, no need for the static_cast here. Updated the pull request accordingly.

PR #650, opened 2026-04-19
Build: Try even harder to get rid of zlib.dll in windows build

insunaa:
## 🍰 Pullrequest

zlib.dll has been a very persistent cockroach in windows builds and this PR attempts to get rid of it once and for all


killerwife:
Just a q, is this aimed at ppl who are starting clean, or upgrading from existing old setup

insunaa:
> Just a q, is this aimed at ppl who are starting clean, or upgrading from existing old setup

Haven't seen the question until today, sorry. It's aimed at both. zlib did something upstream that causes this mess of zlib *always* building as shared, for no discernible reason

PR #676, opened 2026-08-22
ICC: restore Gunship Battle

T-imothy:
Restores the Gunship Battle startup for both factions, cannons and heat, artillery/rocket hull damage, boarding, Below Zero, and the transport and vehicle handling required by the fight.

RP and combat dialogue use BroadcastText rather than the old ScriptText path.

DB companion: https://github.com/cmangos/wotlk-db/pull/1391

Tested as part of complete 10-player and 25-player normal runs. Heroic still needs separate validation.


T-imothy:
Retested the Gunship changes on 10- and 25-player normal with the updated DB companion. Ship damage remains applied instead of regenerating, the cannons overheat and recover normally, and the encounter completes in both modes. Heroic still needs testing.

DB update: cmangos/wotlk-db#1391 at `a8f305cc`.

PR #689, opened 2026-08-24
chore(loot): rename MAX_NR_LOOT_ITEMS definition

natrist:
Rename macro MAX_NR_LOOT_ITEMS to shorter MAX_LOOT_ITEMS.

### DIFF
--- a/src/game/Loot/LootMgr.cpp
+++ b/src/game/Loot/LootMgr.cpp
@@ -996,7 +996,7 @@ GroupLootRoll* Loot::GetRollForSlot(uint32 itemSlot)
 // Inserts the item into the loot (called by LootTemplate processors)
 void Loot::AddItem(LootStoreItem const& item)
 {
-    if (m_lootItems.size() < MAX_NR_LOOT_ITEMS)                              // Normal drop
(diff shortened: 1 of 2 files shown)


killerwife:
Purpose? I generally dont rip git history for no reason. Is this a leaked name?

natrist:
> Purpose? I generally dont rip git history for no reason. Is this a leaked name?

The proposed name is based on the original name in the WoW source code. It's also shorter and more readable.

killerwife:
Ok gimme a source for that name then. Where to find it in client, in what it was leaked, etc.

natrist:
> Ok gimme a source for that name then. Where to find it in client, in what it was leaked, etc.

Screenshot 2026-08-27 at 8 58 25 PM

WoWAE 0.5.3.3368

Also worth noting that in build 12340 the constant is 18 and not 16, as shown in the same function below:
Screenshot 2026-08-27 at 9 07 57 PM

Additional proof here:
Screenshot 2026-08-27 at 9 11 23 PM


I will keep this PR scoped to just the name change and send a follow-up for the value change, if that's alright with you.

PR #698, opened 2026-09-19
ICC: restore Gunship Battle

T-imothy:
Restores Gunship Battle from our working local baseline: crews and boarding waves, cannons/Below Zero, rocket packs, victory, wipe recovery, and progression to Deathbringer's Rise.

Includes the supporting transport and nested-vehicle fixes. Excludes ManTech-only map partitioning and unrelated ICC encounters. Replacement for closed #676.

Validation: 20 C++ translation units passed MSVC syntax/type checks. Gameplay was tested on our full baseline; this upstream-only checkout has not been deployed for a separate gameplay test. Companion DB: https://github.com/cmangos/wotlk-db/pull/1407.


src/game/Entities/Object.cpp
```diff
@@ -458,7 +458,17 @@ void Object::BuildMovementUpdate(ByteBuffer* data, uint16 updateFlags) const
...
     {
         *data << uint32(((Unit*)this)->GetVehicleInfo()->GetVehicleEntry()->m_ID); // vehicle id
-        *data << float(((WorldObject*)this)->GetOrientation());
+
+        // A vehicle carried by a moving transport is oriented in the
+        // transport's local coordinate system. Sending its world orientation
+        // here gives the client the transport rotation twice when the vehicle
+        // becomes the active mover. This is especially visible on gunship
+        // cannons as inverted steering and 180-degree heading flips.
+        WorldObject const* vehicle = static_cast<WorldObject const*>(this);
```
killerwife, line 462: Try GetLocalOrientation()

src/game/Entities/Transports.h
```diff
@@ -113,6 +114,8 @@ class Transport : public GenericTransport
         void SpawnPassengers() override;
         void SpawnPassengersIfDespawned() override;
         void DespawnPassengers() override;
+        void RemoveFromMap();
```
killerwife: The functions are called RemoveFromWorld in cmangos and they are called automatically when an object is being removed

src/game/Entities/Transports.cpp
```diff
@@ -285,6 +285,57 @@ void Transport::DespawnPassengers()
...
+    UpdateForMap(map, false);
+    map->RemoveTransport(this);
+    CleanupsBeforeDelete();
+    ResetMap();
+    delete this;
+}
+
+void Transport::StartMovementNow()
+{
+    SetGoState(GO_STATE_ACTIVE);
```
killerwife: This protocol right here, is based on whats being sent in sniff. Whole GO code assumes nothing else. Make a video perhaps of what is happening without this, because when I was testing this (i added all the gunship chat commands) it always started immediately.

src/game/Entities/Vehicle.cpp
```diff
@@ -943,7 +954,15 @@ void VehicleInfo::RemoveSeatMods(Unit* passenger, uint32 seatFlags)
         if (seatFlags & SEAT_FLAG_CAN_CONTROL)
         {
             CharmInfo* charmInfo = pVehicle->GetCharmInfo();
-            MANGOS_ASSERT(charmInfo);
+
```
killerwife, line 957: Whats the stack trace of this happening, cos the charm info should be cleared on leaving map, rather than arriving at destination. If we do this, we will be deferring other issues elsewhere, cos without the object we cannot do a proper cleanup on player end of the equation since it stores important info like previous flags and stuff. If this was the solution, mind control and charm would also crash, but it doesnt.

src/game/Entities/Transports.cpp
```diff
@@ -285,6 +285,57 @@ void Transport::DespawnPassengers()
...
+
+    if (!GetGOInfo()->moTransport.canBeStopped || !m_currentFrame->IsStopFrame())
+        return;
+
+    // SetGoState clears the manual stop flag, but the normal update still
```
killerwife: TC doesnt do this either.

src/game/MotionGenerators/MovementHandler.cpp
```diff
@@ -896,6 +901,22 @@ bool WorldSession::ProcessMovementInfo(MovementInfo& movementInfo, Unit* mover,
     if (plMover && plMover->IsBeingTeleported() && recv_data.GetOpcode() != CMSG_MOVE_SET_COLLISION_HGT_ACK)
         return false;
 
+    // ICC gunship cannons are rooted transport passengers, but their aim is
```
killerwife: This is not the correct place to do this. If indeed VEHICLE_FLAG_FIXED_POSITION is doing this, client has that information, and would send those flags properly. If there is an issue, its elsewhere. Likely the vehicle is sending wrong flags on spawn or something similar. Check sniff maybe, if not ill have to check it out

src/game/Spells/SpellHandler.cpp
```diff
@@ -123,6 +123,14 @@ void WorldSession::HandleUseItemOpcode(WorldPacket& recvPacket)
     {
         for (const auto& Spell : proto->Spells)
         {
+            // Only spells which are actually cast by using the item can
```
killerwife: No need for the comment in core

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/icecrown_citadel.cpp
```diff
@@ -101,6 +103,24 @@ enum
     SPELL_PUTRICIDE_GAS_VARIABLE     = 74119,
 };
 
+namespace
+{
+Transport* GetGunshipTransport(Map* map, uint32 entry)
```
killerwife, line 108: No need for this, check instance_icecrown_citadel::ExecuteChatCommand on how to fetch them quickly.

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/icecrown_citadel.cpp
```diff
@@ -238,6 +264,27 @@ void instance_icecrown_citadel::OnPlayerEnter(Player* pPlayer)
...
+        pPlayer->DestroyItemCount(ITEM_GOBLIN_ROCKET_PACK,
+            pPlayer->GetItemCount(ITEM_GOBLIN_ROCKET_PACK), true);
+}
+
+void instance_icecrown_citadel::OnPlayerLeave(Player* pPlayer)
+{
+    if (!pPlayer)
+        return;
+
+    // The Goblin Rocket Pack is an ICC Gunship encounter tool. Retail keeps
```
killerwife: Use the dummy aura on the backpack to do this. It has remove flag leave world, and on unapply you can remove the item.

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/icecrown_citadel.cpp
```diff
@@ -680,8 +727,22 @@ void instance_icecrown_citadel::SetData(uint32 uiType, uint32 uiData)
             m_auiEncounter[uiType] = uiData;
             if (uiData == DONE)
             {
-                // enable loot; exact GO entry is handled on object create
-                DoToggleGameObjectFlags(m_uiTeam == ALLIANCE ? GO_GUNSHIP_ARMORY_A : GO_GUNSHIP_ARMORY_H, GO_FLAG_NO_INTERACT, false);
+                // Release controlled cannon riders before moving them away
+                // from the transport.  Teleporting during the same update
+                // can race the vehicle exit packet and leave the client
```
killerwife: Just remove the aura instead of this.

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/icecrown_citadel.cpp
```diff
@@ -710,39 +770,40 @@ void instance_icecrown_citadel::SetData(uint32 uiType, uint32 uiData)
...
                 if (Creature* pSource = GetSingleCreatureFromStorage(m_uiTeam == ALLIANCE ? NPC_GUNSHIP_MURADIN : NPC_GUNSHIP_SAURFANG))
                 {
-                    DoScriptText(m_uiTeam == ALLIANCE ? SAY_GUNSHIP_ALLY_WIN : SAY_GUNSHIP_HORDE_WIN, pSource);
+                    DoBroadcastText(m_uiTeam == ALLIANCE ? SAY_GUNSHIP_ALLY_WIN : SAY_GUNSHIP_HORDE_WIN, pSource);
                     pSource->PlayMusic(0);
-
-                    // ToDo: start WP movement on the deck and prepare to summon the entry NPCs for Saurfang
                 }
 
+                if (Creature* pEnemyCaptain = GetSingleCreatureFromStorage(m_uiTeam == ALLIANCE ? NPC_GUNSHIP_SAURFANG : NPC_GUNSHIP_MURADIN))
```
killerwife: enemyCaptain

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/icecrown_citadel.cpp
```diff
@@ -1005,28 +1092,41 @@ void instance_icecrown_citadel::JustDidDialogueStep(int32 iEntry)
     switch (iEntry)
     {
         case SAY_GUNSHIP_START_ALLY_5:
-            if (GenericTransport* gunship = instance->GetTransport(ObjectGuid(HIGHGUID_MO_TRANSPORT, uint32(GO_ORGRIMS_HAMMER_A))))
-                gunship->SetGoState(GO_STATE_ACTIVE);
-            SetData(TYPE_GUNSHIP_BATTLE, IN_PROGRESS);
+            StartGunshipTransport(instance, GO_ORGRIMS_HAMMER_A);
```
killerwife, line 1068: This relates to the sniff protocol I mentioned.

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/icecrown_citadel.cpp
```diff
@@ -1121,6 +1221,113 @@ void instance_icecrown_citadel::Update(uint32 uiDiff)
...
+                if (Player* player = playerRef.getSource())
+                {
+                    if (player->IsBoarded())
+                        player->ExitVehicle();
+                    player->DestroyItemCount(ITEM_GOBLIN_ROCKET_PACK,
+                        player->GetItemCount(ITEM_GOBLIN_ROCKET_PACK), true);
+                    // This is the destination of the retail victory spell.
+                    // Direct relocation occurs only after the vehicle exit
+                    // has been processed, preventing a stale cannon mover.
+                    player->TeleportTo(instance->GetId(), -548.983f, 2211.24f, 539.29f, 0.0f);
```
killerwife: Is it a true teleport or is it incorporated into the vehicle dismount. Cos vehicle seats are able to do that, send u directly somewhere.

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/gunship_battle.cpp
```diff
@@ -52,12 +67,986 @@ enum
...
+        if (m_wipeTimer > diff)
+            m_wipeTimer -= diff;
+        else
+            m_instance->SetData(TYPE_GUNSHIP_BATTLE, FAIL);
+    }
+};
+
+UnitAI* GetAI_npc_gunship(Creature* creature)
+{
+    return new npc_gunshipAI(creature);
```
killerwife: GetNewAiInstance

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/gunship_battle.cpp
```diff
@@ -52,12 +67,986 @@ enum
...
+        Reset();
+    }
+
+    void Reset() override
+    {
+        // A gunship cannon's ENERGY resource represents stored Heat. Generic
+        // WotLK energy users initialize at 100, but retail cannons begin each
+        // attempt cold and build Heat only through Cannon Blast.
+        m_creature->SetPower(POWER_ENERGY, 0);
+        m_creature->SetImmobilizedState(true);
```
killerwife: Use flag Sessile in DB

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/gunship_battle.cpp
```diff
@@ -52,12 +67,986 @@ enum
...
+    }
+
+    void OnPassengerRide(Unit* /*passenger*/, bool boarded, uint8 /*seat*/) override
+    {
+        if (!boarded)
+        {
+            // The client legitimately reports ROOT for these turrets, but the
+            // generic movement validation can clear it while controlled.
```
killerwife: This must never happen. If this is happening, something else is wrong.

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/gunship_battle.cpp
```diff
@@ -52,12 +67,986 @@ enum
...
+            m_creature->SetImmobilizedState(false);
```
killerwife: Because these lines send packets, and i assume they arent sent in sniff

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/gunship_battle.cpp
```diff
@@ -52,12 +67,986 @@ enum
...
+struct npc_gunship_captainAI : public ScriptedAI
+{
+    npc_gunship_captainAI(Creature* creature) : ScriptedAI(creature),
+        m_instance(static_cast<instance_icecrown_citadel*>(creature->GetInstanceData()))
+    {
+        Reset();
+    }
+
+    instance_icecrown_citadel* m_instance;
+    uint32 m_boardingTimer;
```
killerwife: Port it to timers

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/gunship_battle.cpp
```diff
@@ -52,12 +67,986 @@ enum
...
+        // avoids teleporting or double-transforming the passenger.
+        mage->SetActiveObjectState(true);
+        Movement::MoveSplineInit movement(*mage);
+        movement.MoveTo(positions[0].x, positions[0].y, positions[0].z, false);
+        movement.SetWalk(false);
+        movement.SetFacing(positions[0].o);
+        int32 travelTime = movement.Launch();
+
+        m_freezeMageGuid = mage->GetObjectGuid();
+        mage->AI()->SendAIEvent(AI_EVENT_CUSTOM_A, m_creature, mage,
```
killerwife: Put these on one line instead of two

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/gunship_battle.cpp
```diff
@@ -52,12 +67,986 @@ enum
...
+        DoBroadcastText(m_creature->GetEntry() == NPC_GUNSHIP_MURADIN ? SAY_MURADIN_AGGRO : SAY_SAURFANG_AGGRO, m_creature);
+        m_creature->CastSpell(m_creature, SPELL_CAPTAIN_BATTLE_FURY, TRIGGERED_OLD_TRIGGERED);
+    }
+
+    void EnterEvadeMode() override
+    {
+        if (!m_creature->IsAlive())
+            return;
+
+        // The enemy commander is both a combatant and the controller for the
```
killerwife: If this is an issue, then handle it in instance script, rather than here. You do not need to keep state inside enemy commander.

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/gunship_battle.cpp
```diff
@@ -52,12 +67,986 @@ enum
...
+        {
+            // The commander remains at his station, but must engage players
+            // who land on his gunship to kill the Below Zero mage.
+            if (m_active && IsEnemyCaptain())
+                if (Player* player = SelectGunshipPlayer(m_creature, true))
+                    AttackStart(player);
+            return;
+        }
+
+        if (m_cleaveTimer <= diff)
```
killerwife: These spells can be easily ported to spell lists. They seem easy enough. For example four horsemen have this attack at range when noone is tanking them too

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/gunship_battle.cpp
```diff
@@ -143,8 +1132,13 @@ struct spell_incinerating_blast : public SpellScript
         {
             uint32 damage = spell->GetDamage();
             uint32 energy = caster->GetPower(caster->GetPowerType());
+            // Incinerating Blast consumes 10 Heat before its damage snapshot.
+            // Both TrinityCore and AzerothCore's 3.3.5 implementations retain
```
killerwife: Clear these comments after about other projects

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/gunship_battle.cpp
```diff
@@ -153,19 +1147,186 @@ struct spell_incinerating_blast : public SpellScript
...
+            return false;
+
+        // Freeze the player faction's cannons.  Transport passenger links can
+        // briefly be absent while a moving ship updates its passengers; using
+        // those links here caused only one of the two cannons to receive Below
+        // Zero on otherwise healthy attempts.  The instance team is the stable
+        // encounter authority and also prevents the enemy decorative cannons
+        // from becoming selectable/frozen.
+        instance_icecrown_citadel* instance =
+            static_cast<instance_icecrown_citadel*>(caster->GetInstanceData());
```
killerwife: Need to do a dynamic cast if you want to do existence check

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/gunship_battle.cpp
```diff
@@ -153,19 +1147,186 @@ struct spell_incinerating_blast : public SpellScript
...
+        Unit* target = spell->GetUnitTarget();
+        if (!target || (target->GetEntry() != NPC_ALLIANCE_GUNSHIP_CANNON &&
+                target->GetEntry() != NPC_HORDE_GUNSHIP_CANNON))
+            return;
+
+        // Below Zero's DBC targeting determines which hostile cannons are hit.
+        // Preserve the client spell, then explicitly eject occupied seats. The
+        // generic Eject All Passengers spell is not reliable for cannons that
+        // are passengers of a moving gunship transport.
+        target->SetPower(target->GetPowerType(), 0);
```
killerwife, line 1055: The starting at 0 likely needs to be reflected in some flags. For example i bet the vehicle has no flag for regenerate power, and those ones need to start at 0, like the mana rays in tbc

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/gunship_battle.cpp
```diff
@@ -153,19 +1147,186 @@ struct spell_incinerating_blast : public SpellScript
...
+        if (caster->GetPower(caster->GetPowerType()) >= 100)
+            caster->CastSpell(caster, SPELL_CANNON_OVERHEAT, TRIGGERED_OLD_TRIGGERED);
+    }
+};
+
+/*######
+## spell_gunship_below_zero - 69705
+######*/
+
+struct spell_gunship_below_zero : public SpellScript
```
killerwife: GunshipBelowZero

killerwife (changes requested):

src/game/Entities/Object.cpp
T-imothy: Changed this to GetPosition(vehicle->GetTransport()).o. GetLocalOrientation() is on TransportInfo here, while these cannons use the moving transport. This gets the orientation relative to that transport without the extra branch.

src/game/Entities/Transports.h
T-imothy: Changed this to RemoveFromWorld and the normal removal queue. Removed the separate RemoveFromMap helper and delete-this path.

src/game/Entities/Transports.cpp
T-imothy: Removed the forced path-progress change and went back to the normal SetGoState start. Just played through with those changes and the ships worked, so I am not seeing the old start problem now.

src/game/Entities/Vehicle.cpp
T-imothy: I do not have the old stack trace to give you. I removed the null check workaround and put the assertion back. What I found in the code was BreakCharmOutgoing could reach Uncharm without removing the control-vehicle aura first. It now removes that aura before the fallback, so normal unboarding still has the CharmInfo it needs. Freeze/ejection and getting back into the cannon worked in the latest run.

src/game/MotionGenerators/MovementHandler.cpp
T-imothy: Removed the cannon-specific movement flag changes from here, along with the scripted unroot/root toggling. The cannons now use the DB static flags, including Sessile and no automatic regen, and their existing vehicle flags.

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/icecrown_citadel.cpp
T-imothy: Switched this to Map::GetTransport. I kept a small lookup to get the template counter because that is what Transport::Create uses for the GUID in this checkout, rather than the gameobject entry.

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/icecrown_citadel.cpp
T-imothy: Changed it to remove the item when the on-equip dummy aura is removed. The item uses 73077 for that aura; I checked item 49278 rather than using the separate Rocket Pack Useable spell.

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/icecrown_citadel.cpp
T-imothy: The capture shows the ship casting 72340, Teleport Players on Victory, so I changed this to use that spell and its existing DB destination instead of the hardcoded teleport and delay. That is packet 531343 in the 3.4.3 capture you sent, not an original 3.3.5 capture.

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/gunship_battle.cpp
T-imothy: Moved these to TimerManager. The boarding waves, crew replacements and mage timing now live in the instance rather than on the enemy captain.

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/gunship_battle.cpp
T-imothy: Moved the captain cleave/rending throw and sergeant bladestorm/wounding strike into spell lists. Rending Throw uses the current-target condition and the out-of-melee check now.

src/game/AI/ScriptDevAI/scripts/northrend/icecrown_citadel/icecrown_citadel/gunship_battle.cpp
T-imothy: Added NO_AUTOMATIC_REGEN and disabled power regen in the DB, including the difficulty entries. Removed the scripted SetPower(0); initial power now follows the no-regen flag. Kept the existing Heat Drain aura, and added Suppress Whispers on boarding from the snippet you sent.

T-imothy:
Thanks for going through this. Pushed the review changes and replied on the questions above. Also changed the AI registration, Below Zero name/cast, and cleaned up the formatting and comments you pointed out. Just played through with these changes on my playerbot-enabled baseline: cannons, axe throwers, rockets, mage freeze/ejection and getting back in, portal reavers, Muradin helping fight, and the loot cache all worked. The DB changes are up in https://github.com/cmangos/wotlk-db/pull/1407 too.

T-imothy:
Pushed a follow-up for issues we found after the review changes. Frozen cannons now stay out of normal NPC evade handling, and the crews use the actual ship when choosing targets, including when someone enters or leaves a cannon. Tested the latest version in both 10 and 25 and the fight worked. There were still some delays with boarders engaging and Muradin returning home, so I am not calling those timings verified.

sql/base/dbc/cmangos_fixes/Spell.sql
```diff
@@ -27,6 +27,19 @@ INSERT INTO `spell_template` (`Id`,`Attributes`,`AttributesEx`,`AttributesEx2`,`
...
 ('62388','0','0','0','0','0','101','21','6','1','0','0','4','0','0','0','1','Aura required for Demonic Circle 48020');
 
+
+-- Zafod casts 70055 in the 3.4.3 ICC capture (2024-01-23, packet 500035).
+-- The ID is absent from 3.3.5 client data; create-item fields are reconstructed.
+INSERT INTO `spell_template`
+    (`Id`, `Attributes`, `AttributesEx2`, `ProcChance`, `Effect1`,
+     `EffectImplicitTargetA1`, `EffectItemType1`, `SpellName`, `SchoolMask`)
+VALUES
+    (70055, 384, 4, 101, 24, 25, 49278, 'Create Rocket Pack', 1)
```
killerwife: Do it in one line and without the on duplicate key. That should never happen.

sql/scriptdev2/spell.sql
```diff
@@ -2463,3 +2463,8 @@ INSERT INTO spell_scripts(Id, ScriptName) VALUES
 (26505, 'spell_lunar_festival_cluster_big_u'),
 (26510, 'spell_lunar_festival_cluster_big_u'),
 (26515, 'spell_lunar_festival_cluster_big_u');
+
+DELETE FROM spell_scripts WHERE Id IN (73077,72340);
```
killerwife: Add it to the wotlk section without the DELETE. The insert should always go through without issue

